Issue #2, opened 2017-04-02
Encoding support


fetzerms:
The module description states, that this is yEnc **Encoding**/Decoding for Python. But I do not see any Decoding functions offered by the header file (nor by the source). 

Is this a feature that will be added? Or is this module aimed purely at sabnzbd, and encoding/posting support is not planned?

Safihre:
You are absolutely right: there is no encoding support and there also won't be (at least, I won't make it, but other are free to add it of course).
In a next iteration I will remove any references to Encoding!

fetzerms:
Ok. I suggest to keep this issue open and close it, when the references have been removed.

PR #3, opened 2017-04-02
Removed references to encoding support. See Issue #2

fetzerms:
See title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-SABYenc - yEnc Encoding/Decoding for Python
+SABYenc - yEnc Decoding for Python
 ===============================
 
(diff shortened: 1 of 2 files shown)


Safihre:
Thanks but I can't merge it now because of the linked build system that will auto submit the new 2.9.0 release to pypi. We don't want that yet because sabnzbd itself expects 2.8.0.
Its a bit clumsy, but I will merge this when we want 2.9.0 out :)

fetzerms:
Ok, thanks for the info. You can simply close the PR, if you want to. Or you can leave it open and merge whenever the time is right. :+1:

PR #7, opened 2018-01-03
add pytest config skipping scanning of yencfiles

jsbronder:
Somewhere between pytest-3.0.3 and 3.2.2, the txt files in yencfiles
started to get picked up and scanned which led to a bunch of errors
like:

```
================================================================ ERRORS ================================================================
___________________________________________ ERROR collecting tests/yencfiles/test_badcrc.txt ___________________________________________
/usr/lib64/python2.7/site-packages/py/_path/common.py:162: in read_text
    return f.read()
/usr/lib64/python2.7/codecs.py:314: in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
E   UnicodeDecodeError: 'utf8' codec can't decode byte 0xa1 in position 189: invalid start byte
```

While there may be a more clever fix, those files shouldn't be getting
scanned anyways.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+norecursedirs = yencfiles


jsbronder:
Gentoo bug report https://bugs.gentoo.org/show_bug.cgi?id=641820

Safihre:
Oh wow, so that's why I was getting these stupid errors!
I just downgraded to an old version of `pytest`.

Thanks!

PR #20, opened 2022-02-22
Integrate node-yencode

animetosho:
This pull request replaces the encoder/decoder/CRC code in Sabyenc with the SIMD implementations from https://github.com/animetosho/node-yencode

Discussion follows from https://github.com/animetosho/node-yencode/issues/11  
See above thread for notes on breaking test cases and other suggestions.

I believe work on this is complete enough for review/merging.  
A remaining issue is how to accurately determine the build architecture, as I don't know anything better than Python's `platform.machine()` which [has issues](https://github.com/animetosho/node-yencode/issues/11#issuecomment-1046850475).

During testing, I noticed that `PyBytes_FromStringAndSize` can take a fair bit of time (particularly on MSVC/Windows). You can eek out more perf if you use pre-allocated buffers and write directly into that, rather than create + copy into a new buffer every time.


setup.py
```diff
@@ -22,10 +22,140 @@
...
 
+BUILD_DIR = "build"
+machine = platform.machine().lower()  # TODO: this is unreliable, need something better to detect target architecture
+is_armv7 = (machine.startswith("armv7"))  # TODO: can armv8_ still be AArch32?
+is_arm = (is_armv7 or machine.startswith("arm") or machine.startswith("aarch64"))
+is_x86 = (machine in ["i386", "i686", "x86", "x86_64", "x64", "amd64"])
+
+
+
+class sabyenc_build(build_ext):
```
Safihre, line 40: Nice, you already found it! I was just about to write that this seems indeed be the way to implement any custom build work.

Safihre (commented):
Should we already work on using buffers in this PR?
Or just merge this and go from there?
This PR is quite large because of the `crcutil` stuff.

setup.py
```diff
@@ -22,10 +22,140 @@
...
+from setuptools.command.build_ext import build_ext
+import sysconfig
+import platform
 
 with open("README.md", "r") as file_long_description:
     long_description = file_long_description.read()
 
+BUILD_DIR = "build"
+machine = platform.machine().lower()  # TODO: this is unreliable, need something better to detect target architecture
+is_armv7 = (machine.startswith("armv7"))  # TODO: can armv8_ still be AArch32?
```
Safihre, line 34: My NAS has a  ARMv8  and `platform.machine` reports `aarch64`.
So it doesn't include the NEON stuff?

animetosho: AArch64 applies NEON by default (and `-mfpu=neon` isn't recognised).

animetosho:
> On aarch64 / ARM64, I get this:

I presume current Sabyenc master works fine there?  
Losing `sabyenc3.encode` doesn't seem like something this PR should affect...

I otherwise don't know why you'd be getting that. It works on my AArch64 platform.

> deleted this message about RISCV64, because I can reproduce on x86:

The old CRC specified in the test was 0xE47AE66A / 3833259626, which is incorrect:

	python3 -c "import binascii; print(binascii.crc32(b'Hello world!'))"
	461707669

I fixed the test to use the correct CRC, which the new code should be generating. Are you sure you're testing this with this PR's version?

sanderjo:
> > On aarch64 / ARM64, I get this:
> 
> I presume current Sabyenc master works fine there? Losing `sabyenc3.encode` doesn't seem like something this PR should affect...

Ouch ... happens on @Safihre safihre 's sabyenc too ... so https://github.com/sabnzbd/sabyenc/issues/21

animetosho:
Ah, good to know.  
I pushed a change to see if it'd make a difference - change is fine to keep as is, regardless (it should preserve original behaviour slightly better).

sanderjo:
FWIW

To verify there is NEON stuff in my Aarch64 sabyenc, I did some ugly stuff:
```

sander@nanopineo2:~/git/animetosho_sabyenc$ objdump -d build/lib.linux-aarch64-3.6/sabyenc3.cpython-36m-aarch64-linux-gnu.so | grep -i neon | wc -l
227
```

and even more hacky:


```
sander@nanopineo2:~/git/animetosho_sabyenc$ objdump -d build/lib.linux-aarch64-3.6/[sabyenc3.cpython-36m-aarch64-linux-gnu.so](http://sabyenc3.cpython-36m-aarch64-linux-gnu.so/) | grep -i ", v"  | head
    3920: 4e211f39 and v25.16b, v25.16b, v1.16b
    3924: 4e211f7b and v27.16b, v27.16b, v1.16b
    3928: 4e211f18 and v24.16b, v24.16b, v1.16b
    392c: 4e39bf18 addp v24.16b, v24.16b, v25.16b
    3930: 4e211e79 and v25.16b, v19.16b, v1.16b
    3934: 4e211c00 and v0.16b, v0.16b, v1.16b
    3938: 4e3bbc00 addp v0.16b, v0.16b, v27.16b
    393c: 4e211efb and v27.16b, v23.16b, v1.16b
    3940: 4e38bc00 addp v0.16b, v0.16b, v24.16b
    3944: 4e3bbf38 addp v24.16b, v25.16b, v27.16b

```

Looks good (if my checks are correct)

sanderjo:
And about the same trick for SIMD SSE on my x86:

```
sander@witte2004:~/git/animetosho_sabyenc$ objdump -d /home/sander/git/animetosho_sabyenc/build/lib.linux-x86_64-3.8/sabyenc3.cpython-38-x86_64-linux-gnu.so | grep -i SSE | grep -vi assembly | grep -vi classes | wc -l
1631
```

... and no SSE / SIMD in the old sabyenc3.so:
```
sander@witte2004:~/git/animetosho_sabyenc$ objdump -d /home/sander/.local/lib/python3.8/site-packages/sabyenc3.cpython-38-x86_64-linux-gnu.so | grep -i SSE | grep -vi assembly | grep -vi classes | wc -l
0
```

Looks good.

animetosho:
If you have `perf` installed (package named `linux-perf` in Debian), you can see what's being run the most using

	perf record python3 tests/speed_compare.py
	perf report

The top function should be the selected yEnc decoder.

setup.py
Safihre: Check, good to know.. We will probably have to introduce some support for command line parameters to specify custom target arch so we can support cross compiling.
Have go check how nzbget fixed this in his universal Linux installer.

animetosho: Getting the arch is dependent on the build system, so you can't really take a solution from a different build system and use it here.  
Ideally, CCompiler would have some sort of 'get_arch' function, as the compiler would know what it's targeting.

It doesn't look like there's much support for cross compiling in setuptools. It seems like it has some support for MSVC, but otherwise, examples I've seen seem to mostly override the selected compiler with a cross compiler, rather than use something built into setuptools.

I did come across `distutils.util.get_platform()`, which looks better, but don't know how definitive it is.

Overall, from my quick look, setuptools doesn't look like an extensive build system, so you may have to live with its limitations.

Safihre:
Just so I understand correctly: we can compile it on a system that has all x86 features (AVX, SSSE, etc) and then it only uses the supported one on the system at runtime?
Because of course to use it in SABnzbd we need to pre-compile it for use in the executable.

Safihre:
@animetosho To structure a bit the questions and next steps:
- First of all, thank you very much for all this work, it's much more than I was hoping for! 
- They way it is setup, can it be compiled on a system that has all x86 SIMD features (AVX, SSSE, etc) and then it only uses the supported SIMD on the system at runtime? I thought there was runtime detection in there, right?
- Shall I just merge and then do some code-cleanup to get it all in the same style?
- I will try to see if GitHub Actions can build for us.
- The switch to using Python buffers, I can try to work on this myself, it seems easy enough.

animetosho:
> They way it is setup, can it be compiled on a system that has all x86 SIMD features (AVX, SSSE, etc) and then it only uses the supported SIMD on the system at runtime? I thought there was runtime detection in there, right?

The build system (hardware) doesn't need to support any of the SIMD features, the *compiler* (software) does. I've only enabled up to AVX2, which was first released in hardware in 2013 (software support should be earlier), so most compilers should support all of them.  
Should be a similar story for ARM.

Yes, there's dynamic dispatch, so it'll only use what's available on the running system (build system is irrelevant). Obviously you won't be able to *test* the kernels if the hardware doesn't support them though.  
If you want to double check this, and you don't have access to older hardware, some VMs allow you to change CPUID, so CPU features can be masked out. If you don't have newer hardware, functionality can be tested in an emulator (probably qemu or Intel's Software Development Emulator).

> Shall I just merge and then do some code-cleanup to get it all in the same style?

Do whatever works best for you.

Glad to be of help!

Safihre:
@animetosho Tried to also build a Windows 32bit release (since we also have a SABnzbd 32bit version) for some ancient computers out there. But for some reason it fails:
https://github.com/sabnzbd/sabyenc/runs/5304588371?check_suite_focus=true

sanderjo:
A big step for sabyenc. Ah ... SAByenc version has jumped to 5.0.0 ... deserved!

@animetosho Thank you!

sanderjo:
Oh, more impressive results (which I had not expected):

On my RISCV64 system (so: not X86, not ARM, so no SIMD instructions implemented), sabyenc 5.0.0 is more than twice as fast as sabyenc 4.0.2. Duration drops from 5600 ms to 2300 ms. 

```
sipeed@sipeed:~/git/poging2$ for i in `seq 1 5`; do tests/speed_compare.py; done
4.0.2
---
     yEnc C New  took  5597 ms
---
4.0.2
---
     yEnc C New  took  5597 ms
---
4.0.2
---
     yEnc C New  took  5601 ms
---
4.0.2
---
     yEnc C New  took  5601 ms
---
4.0.2
---
     yEnc C New  took  5600 ms
---
```

```
sipeed@sipeed:~/git/poging2$ cp build/lib*/*so tests/
sipeed@sipeed:~/git/poging2$ for i in `seq 1 5`; do tests/speed_compare.py; done
5.0.0
---
     yEnc C New  took  2291 ms
---
5.0.0
---
     yEnc C New  took  2293 ms
---
5.0.0
---
     yEnc C New  took  2291 ms
---
5.0.0
---
     yEnc C New  took  2292 ms
---
5.0.0
---
     yEnc C New  took  2291 ms
---
```

sanderjo:
FYI: Results on an ARM64:

Duration of 5100 ms drops to 830 ms

(A bit strange: this ARM64 feels much faster & responsive than RISCV64, but the sabyenc 4.0.2 shows the same performance. Anyway)

```
sander@nanopineo2:~/git/sabyenc$ for i in `seq 1 5`; do tests/speed_compare.py; done
4.0.1
---
     yEnc C New  took  5144 ms
---
4.0.1
---
     yEnc C New  took  5141 ms
---
4.0.1
---
     yEnc C New  took  5161 ms
---
4.0.1
---
     yEnc C New  took  5135 ms
---
4.0.1
---
     yEnc C New  took  5143 ms
---
```

and

```
sander@nanopineo2:~/git/sabyenc$ for i in `seq 1 5`; do tests/speed_compare.py; done
5.0.0
---
     yEnc C New  took   801 ms
---
5.0.0
---
     yEnc C New  took   835 ms
---
5.0.0
---
     yEnc C New  took   813 ms
---
5.0.0
---
     yEnc C New  took   804 ms
---
5.0.0
---
     yEnc C New  took   854 ms
---
```

Safihre:
A quick-test on an Apple M1 shows a massive 10-fold performance increase (top is the new version).
image

PR #24, opened 2022-02-25
Updates to integration of yencode

Safihre:
@animetosho Maybe you can take an expert look at my changes?

I found we can indeed create a Python bytes-buffer and directly write to the underlying C-pointer.
It dramatically increases performance on Windows (like you already predicted!) without needing to change the input/output.
On Windows this drops my `speed_compare` times from 950ms with the old `sabyenc3` to 30ms with this version.. Crazy ⏩!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ class SAByEncBuild(build_ext):
                 "-O3",
                 "-fPIC",
                 "-fwrapv",
+                "-std=c++14",
             ]
 
         srcdeps_crc_common = ["yencode/common.h", "yencode/crc_common.h", "yencode/crc.h"]
@@ -71,6 +72,10 @@ class SAByEncBuild(build_ext):
         output_dir = os.path.dirname(self.build_lib)
         compiled_objects = []
         for source_files in [
+            {
+                "sources": ["yencode/platform.cc", "yencode/encoder.cc", "yencode/decoder.cc", "yencode/crc.cc"],
+                "include_dirs": ["crcutil-1.0/code", "crcutil-1.0/examples"],
+            },
             {
                 "sources": ["yencode/encoder_sse2.cc"],
                 "depends": srcdeps_enc_common + ["encoder_sse_base.h"],
@@ -145,7 +150,7 @@ class SAByEncBuild(build_ext):
                     "crcutil-1.0/code/multiword_128_64_gcc_amd64_sse2.cc",
                     "crcutil-1.0/examples/interface.cc",
                 ],
-                "gcc_flags": ["-Wno-expansion-to-defined"],
+                "gcc_flags": ["-Wno-expansion-to-defined", "-Wunused-parameter"],
                 "include_dirs": ["crcutil-1.0/code", "crcutil-1.0/tests"],
                 "macros": [("CRCUTIL_USE_MM_CRC32", "0")],
             },
@@ -189,8 +194,7 @@ setup(
     ext_modules=[
         Extension(
             "sabyenc3",
-            ["src/sabyenc3.c", "yencode/platform.cc", "yencode/encoder.cc", "yencode/decoder.cc", "yencode/crc.cc"],
-            include_dirs=["crcutil-1.0/code", "crcutil-1.0/examples"],
+            ["src/sabyenc3.cc"],
         )
     ],
     cmdclass={"build_ext": SAByEncBuild},
--- a/src/sabyenc3.cc
+++ b/src/sabyenc3.cc
@@ -30,7 +30,7 @@
 /* Function declarations */
 PyMODINIT_FUNC PyInit_sabyenc3(void);
 static size_t decode_buffer_usenet(PyObject *, char *, int, char **, Bool *);
-static char * find_text_in_pylist(PyObject *, char *, char **, int *);
+static char * find_text_in_pylist(PyObject *, const char *, char **, int *);
 int extract_filename_from_pylist(PyObject *, int *, char **, char **, char **);
 uLong extract_int_from_pylist(PyObject *, int *, char **, char **);
 
@@ -292,7 +292,7 @@ static size_t decode_buffer_usenet(PyObject *Py_input_list, char *output_buffer,
     We need a special function to find the keywords
     because they can be split over multiple chunks.
 */
-static char * find_text_in_pylist(PyObject *Py_input_list, char *search_term, char **cur_char, int *cur_index) {
+static char * find_text_in_pylist(PyObject *Py_input_list, const char *search_term, char **cur_char, int *cur_index) {
     // String holders
     char *next_string = NULL;
     char *start_loc = NULL;
@@ -534,7 +534,7 @@ PyObject* decode_usenet_chunks(PyObject* self, PyObject* args) {
     // Resize data to actual value
     // We use this instead of "_PyBytes_Resize", as it seems to cause a drop in performance
     Py_SIZE(sv) = output_len;
-    sv->ob_sval[output_len] = "\0";
+    sv->ob_sval[output_len] = '\0';
     sv->ob_shash = -1;
 
     // Build output


setup.py
```diff
@@ -20,33 +20,36 @@
...
+
+        # Determine compiler flags
+        if self.compiler.compiler_type == "msvc":
+            # LTCG not enabled due to issues seen with code generation where
+            # different ISA extensions are selected for specific files
+            ldflags = ["/OPT:REF", "/OPT:ICF"]
+            cflags = ["/O2", "/GS-", "/GL-", "/Gy", "/sdl-", "/Oy", "/Oi"]
         else:
             # TODO: consider -flto - may require some extra testing
-            base_ldflags = sysconfig.get_config_var("LDFLAGS").split()
```
Safihre, line 49: @animetosho I removed this part, because this picks up the FLAGS that Python itself was compiled with.
On cross-compiled version, this will reference the locations on the Python-build-system, and not the current-system-locations.
Setuptools automatically already adds almost all the parameters below. You can see that for example here, where many parameters are listed twice:
https://github.com/sabnzbd/sabyenc/runs/5332039624?check_suite_focus=true#step:5:354
Just wondering why you added the `sysconfig.get_config_var("LDFLAGS")` part? Maybe I'm missing something :)

Safihre:
Hmm sadly it seems that using the bytes-buffer makes it much faster on Windows, it becomes slower on Linux 😢 
Went from 70ms to 130ms with the pre-allocated PyBytes.

animetosho:
The idea I actually had was Python passing in the buffer, and the function would return the resulting length. This saves allocation + deallocation for each decode call, but obviously changes the API.

For keeping the API the same, it looks fine. Your speed difference does sound odd (I only measured ~40% for the additional copy on Windows). Don't know about the Linux issue - maybe it depends on the realloc implementation? You can try either running a profiler (e.g. `perf`) or remove the resize call to see if it makes a difference.

setup.py
animetosho: Oh nice.

I never checked - [the example](https://pythonextensionpatterns.readthedocs.io/en/latest/compiler_flags.html#setting-flags-automatically-in-setup-py) provided used it, so I went along (since most compilers don't mind flags being duplicated).

Safihre:
Yes indeed, I want to change the API to your suggestion using buffers in a later version, as it requires big changes in SABnzbd itself.

animetosho:
The performance under Linux is curious.

2x alloc + copy, under `perf stat`

	---
	     yEnc C New  took    34 ms
	---
	
	 Performance counter stats for 'python3 tests/speed_compare.py':
	
	             56.38 msec task-clock                #    0.983 CPUs utilized
	                 5      context-switches          #    0.089 K/sec
	                 0      cpu-migrations            #    0.000 K/sec
	             2,137      page-faults               #    0.038 M/sec
	       186,303,972      cycles                    #    3.304 GHz
	       327,250,519      instructions              #    1.76  insn per cycle
	        30,943,463      branches                  #  548.814 M/sec
	         1,147,825      branch-misses             #    3.71% of all branches
	
	       0.057367360 seconds time elapsed
	
	       0.057374000 seconds user
	       0.000000000 seconds sys

alloc + resize

	---
	     yEnc C New  took    78 ms
	---
	
	 Performance counter stats for 'python3 tests/speed_compare.py':
	
	            100.96 msec task-clock                #    0.994 CPUs utilized
	                10      context-switches          #    0.099 K/sec
	                 2      cpu-migrations            #    0.020 K/sec
	            49,165      page-faults               #    0.487 M/sec
	       356,909,360      cycles                    #    3.535 GHz
	       486,050,315      instructions              #    1.36  insn per cycle
	        59,378,406      branches                  #  588.137 M/sec
	         1,263,520      branch-misses             #    2.13% of all branches
	
	       0.101608512 seconds time elapsed
	
	       0.048720000 seconds user
	       0.052780000 seconds sys

So it seems that saving the copy is indeed faster (lower user time), but for some reason, it invokes the kernel making it much slower overall.

If the `_PyBytes_Resize` call is removed, it shows that the issue is with the resize operation:

	---
	     yEnc C New  took    25 ms
	---
	
	 Performance counter stats for 'python3 tests/speed_compare.py':
	
	             46.65 msec task-clock                #    0.977 CPUs utilized
	                 4      context-switches          #    0.086 K/sec
	                 0      cpu-migrations            #    0.000 K/sec
	             2,045      page-faults               #    0.044 M/sec
	       148,497,384      cycles                    #    3.183 GHz
	       326,626,487      instructions              #    2.20  insn per cycle
	        30,812,838      branches                  #  660.556 M/sec
	         1,118,434      branch-misses             #    3.63% of all branches
	
	       0.047732582 seconds time elapsed
	
	       0.039780000 seconds user
	       0.007956000 seconds sys

I don't know why the resize makes things much slower, but maybe it's a libc implementation detail (I'm using glibc 2.31), as I'm guessing `_PyBytes_Resize` eventually calls `realloc`. Maybe I'll test under musl.

I wonder if there's some way to set the size of a buffer without changing its allocation. Otherwise what you have here may be the best solution for a no-API change.

Safihre:
Interesting. The C code for the resize is here: https://github.com/python/cpython/blob/5a1c637ec6264790d3cfeef46815c62c32b510f3/Objects/bytesobject.c#L2989
I specifically checked all the Bytes related C function calls to find the most optimal way, that's how I came up with this.

animetosho:
Looking at the code, it seems like [this](https://github.com/python/cpython/blob/5a1c637ec6264790d3cfeef46815c62c32b510f3/Objects/bytesobject.c#L3032-L3035) is setting the size.

Replacing the resize line in sabyenc3.c with the following seems to work for me:

	PyBytesObject *sv = (PyBytesObject *)Py_output_buffer;
	Py_SET_SIZE(sv, output_len);
	output_buffer[output_len] = '\0';
	sv->ob_shash = -1;          /* invalidate cached hash value */

Whilst the null terminator and `Py_SET_SIZE` [is documented](https://docs.python.org/3/c-api/structures.html#c.Py_SET_SIZE), I'm not sure how much this can be relied upon, since there doesn't seem to be a documented connection between `Py_SIZE` and `PyBytes_Size` (even though they're [the same](https://github.com/python/cpython/blob/5a1c637ec6264790d3cfeef46815c62c32b510f3/Objects/bytesobject.c#L1189)).  
`PyObject_Realloc` also currently [just calls `realloc`](https://github.com/python/cpython/blob/7d8b69e1d1f125454d8cec81ff0dee72f2bef957/Objects/obmalloc.c#L719), but if that behavior changes in the future, it may cause the above to not work (though it seems unlikely, as memory allocators usually track the size themselves).

Safihre:
I applied the changes. Indeed now on all systems performance is good.
So you do not think there's a risk of some memory leaking since we don't call `realloc`?

Safihre:
And it now gives this error, should I worry about it?
```
  src/sabyenc3.c:537:29: warning: assignment to ‘char’ from ‘char *’ makes integer from pointer without a cast [-Wint-conversion]
    537 |     sv->ob_sval[output_len] = "\0";
        |                             ^
```

animetosho:
> So you do not think there's a risk of some memory leaking since we don't call realloc?

There won't be any leak as the memory allocator doesn't care about that size.

If you want to be extra sure, you can write a test case which constantly decodes in a loop (preferably with something that results in large discrepancy between reserved and required bytes, e.g. 1MB of newlines followed by another character) and see if the memory usage goes up.

Safihre:
@animetosho do you think there's any harm in setting specifically `-std=c++14`? It was to remove some compiler warnings on macOS.

animetosho:
The newer versions of C++ are backwards compatible, so there's no issue other than locking out older compilers.  
What's some of the compiler warnings?

Safihre:
It's not showing on the CI, but on my own macOS systems it shows multiple of these:
```
  clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -arch arm64 -arch x86_64 -g -Icrcutil-1.0/code -Icrcutil-1.0/examples -I/Library/Frameworks/Python.framework/Versions/3.10/include/python3.10 -c yencode/encoder.cc -o build/temp.macosx-10.9-universal2-3.10/yencode/encoder.o
  In file included from yencode/encoder.cc:1:
  yencode/common.h:184:12: warning: generalized initializer lists are a C++11 extension [-Wc++11-extensions]
      return {a, b};
             ^~~~~~
  yencode/common.h:187:12: warning: generalized initializer lists are a C++11 extension [-Wc++11-extensions]
      return {a, b};
             ^~~~~~
  yencode/common.h:190:12: warning: generalized initializer lists are a C++11 extension [-Wc++11-extensions]
      return {a, b, c};
             ^~~~~~~~~
  yencode/common.h:193:12: warning: generalized initializer lists are a C++11 extension [-Wc++11-extensions]
      return {a, b, c, d};
```
Just adding `-std=c++11` fixed that.

animetosho:
Interesting - I wonder why the non-Mac Clang doesn't show that.

Tried to [work around it](https://github.com/animetosho/node-yencode/commit/17e4cab8b240d546b9f2212ee88eca7ffba5e501), dunno if it works, though your solution is fine too.  
Thanks for letting me know.

PR #31, opened 2022-03-12
Fix building on OpenBSD

bket:
Building of sabyenc-5.1.0 fails on OpenBSD current (amd64) with:

```
In file included from src/yencode/encoder_sse2.cc:4:
src/yencode/encoder_sse_base.h:33:5: error: use of undeclared identifier 'aligned_alloc'
    ALIGN_ALLOC(lookups, sizeof(*lookups), 16);
    ^
src/yencode/common.h:46:56: note: expanded from macro 'ALIGN_ALLOC'
#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
                                                       ^
1 error generated.
```

With this diff building is successful, and all tests pass.

### DIFF
--- a/src/yencode/common.h
+++ b/src/yencode/common.h
@@ -39,7 +39,7 @@
 #include <stdlib.h> // MSVC ARM64 seems to need this
 #define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = _aligned_malloc((len), align)
 #define ALIGN_FREE _aligned_free
-#elif defined(__cplusplus) && __cplusplus >= 201100 && !(defined(_MSC_VER) && (defined(__clang__) || defined(_M_ARM64) || defined(_M_ARM))) && !defined(__APPLE__)
+#elif defined(__cplusplus) && __cplusplus >= 201100 && !(defined(_MSC_VER) && (defined(__clang__) || defined(_M_ARM64) || defined(_M_ARM))) && !defined(__APPLE__) && !defined(__OpenBSD__)
 // C++11 method
 // len needs to be a multiple of alignment, although it sometimes works if it isn't...
 #include <cstdlib>


Safihre:
@animetosho FYI

animetosho:
Thanks for pointing that out.

I originally thought that `aligned_alloc` was added in C++11, and the Apple Clang compiler was just an oddball, but looking into it more, it's actually only added in C11. The function only got adopted into C++ in C++17.  
So instead of adding more exceptions, perhaps [this block](https://github.com/sabnzbd/sabyenc/blob/be91bb03cca3fcf82f532b68620dcf454a90a488/src/yencode/common.h#L38-L54) could be replaced with

	#include <stdlib.h>
	#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
		// MSVC doesn't support C11 aligned_alloc: https://stackoverflow.com/a/62963007
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = _aligned_malloc((len), align)
		#define ALIGN_FREE _aligned_free
	#elif defined(_ISOC11_SOURCE)
		// C11 method
		// len needs to be a multiple of alignment, although it sometimes works if it isn't...
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
		#define ALIGN_FREE free
	#elif defined(__cplusplus) && __cplusplus >= 201700
		// C++17 method
		#include <cstdlib>
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = std::aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
		#define ALIGN_FREE free
	#elif (defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112) || (defined(_XOPEN_SOURCE) && _XOPEN_SOURCE >= 600)
		#define ALIGN_ALLOC(buf, len, align) if(posix_memalign((void**)&(buf), align, (len))) (buf) = NULL
		#define ALIGN_FREE free
	#else
		#error Could not find available aligned allocator
	#endif

Would you be able to check whether it works for you?  
Will also need to be tested on MacOS.

bket:
@animetosho, on OpenBSD the suggested change initially fails with:

```
In file included from src/yencode/platform.cc:1:
src/yencode/common.h:57:3: error: Could not find available aligned allocator
        #error Could not find available aligned allocator
         ^
1 error generated.

```

Adding either `-D_POSIX_C_SOURCE=200112` or `-D_XOPEN_SOURCE=600` to CFLAGS makes sabyenc build again.

animetosho:
Thanks for trying that! It seems like the `aligned_alloc` detection is working at least; I guess the POSIX detection not so much, so will need to revert that to the previous style, for the odd systems out there.

I guess this should be it then:

	#include <stdlib.h>
	#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
		// MSVC doesn't support C11 aligned_alloc: https://stackoverflow.com/a/62963007
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = _aligned_malloc((len), align)
		#define ALIGN_FREE _aligned_free
	#elif defined(_ISOC11_SOURCE)
		// C11 method
		// len needs to be a multiple of alignment, although it sometimes works if it isn't...
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
		#define ALIGN_FREE free
	#elif defined(__cplusplus) && __cplusplus >= 201700
		// C++17 method
		#include <cstdlib>
		#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = std::aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
		#define ALIGN_FREE free
	#else
		#define ALIGN_ALLOC(buf, len, align) if(posix_memalign((void**)&(buf), align, (len))) (buf) = NULL
		#define ALIGN_FREE free
	#endif

bket:
> #include <stdlib.h>
> #if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
> 	// MSVC doesn't support C11 aligned_alloc: https://stackoverflow.com/a/62963007
> 	#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = _aligned_malloc((len), align)
> 	#define ALIGN_FREE _aligned_free
> #elif defined(_ISOC11_SOURCE)
> 	// C11 method
> 	// len needs to be a multiple of alignment, although it sometimes works if it isn't...
> 	#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
> 	#define ALIGN_FREE free
> #elif defined(__cplusplus) && __cplusplus >= 201700
> 	// C++17 method
> 	#include <cstdlib>
> 	#define ALIGN_ALLOC(buf, len, align) *(void**)&(buf) = std::aligned_alloc(align, ((len) + (align)-1) & ~((align)-1))
> 	#define ALIGN_FREE free
> #else
> 	#define ALIGN_ALLOC(buf, len, align) if(posix_memalign((void**)&(buf), align, (len))) (buf) = NULL
> 	#define ALIGN_FREE free
> #endif

Yes, this works!

Safihre:
Applied the changes. It will build and release `5.1.1` now.

PR #38, opened 2022-03-28
Implement compile-time checks for flags/includes/defines

### LINKED ISSUE #37: Build failures on Synology
I tried building on the SynoCommunity platform for all the Synology platforms and some more stuff came up:
https://github.com/Safihre/spksrc/actions/runs/2027059583

We have a nice selection:
* No support for `c++11`: https://github.com/Safihre/spksrc/runs/5656681505?check_suite_focus=true#step:8:28
  Will try to see if I can dynamically check for `c++11` support in `setup.py` by calling `gcc --help -v`
* `unrecognized argument in option ‘-march=armv8-a+crc’` https://github.com/Safihre/spksrc/runs/5656681460?check_suite_focus=true#step:6:30185
 This is on the `hi3535`, not sure what kind of platform this is. Maybe the fix from #35 fixes this also.
* ~`<built-in>: note:   initializing argument 3 of` https://github.com/Safihre/spksrc/runs/5656681412?check_suite_focus=true#step:6:30441~
 ~Seems same as #35~

Safihre:
@animetosho What do you think about my attempt? It seems to work quite wel!

Only thing remains is the `sys/auxv.h` problem on ARMv5. Just setting `IS_ARM = False` is not enough. 
Probably we need to `#undef PLATFORM_ARM`, right?
Would you do this with external detection from the `setup.py` or just detect it in the C code and manage it there?

Closes #37

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ def autoconf_check(
 
 
 class SAByEncBuild(build_ext):
-    def build_extension(self, ext):
+    def build_extension(self, ext: Extension):
         # Try to determine the architecture to build for
         machine = platform.machine().lower()
         IS_X86 = machine in ["i386", "i686", "x86", "x86_64", "x64", "amd64"]
@@ -111,8 +111,10 @@ class SAByEncBuild(build_ext):
             # gcc before 4.7 called it "-std=c++0x"
             if autoconf_check(self.compiler, flag_check="-std=c++11"):
                 cflags.append("-std=c++11")
+                ext.extra_compile_args.append("-std=c++11")
             elif autoconf_check(self.compiler, flag_check="-std=c++0x"):
                 cflags.append("-std=c++0x")
+                ext.extra_compile_args.append("-std=c++0x")
             else:
                 log.info("C++11 flag not available")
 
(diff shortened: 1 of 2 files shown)


animetosho:
Don't have a lot of time to investigate at the moment, but you might need a check to avoid doing this for Microsoft's compiler.

A main function also isn't needed if you just assemble (`-S` switch) and you can direct the output to /dev/null (`-o /dev/null` switch) to save writing + deleting the output.  
In fact, if you just want to test flags, something like `cc -S -x c++ /dev/null -o /dev/null [flags]` will work.  
(if you want to avoid writing the input files at runtime, you could just put it into the Git repo)

I don't have an ARMv5 platform to test, so don't know how to detect it in code. I thought ARMv5 was quite different from v6, but it looks like the [Godbolt ARM compilers](https://godbolt.org/z/rajM3E49d) report targeting v4/v5 by default, so perhaps there's no easy way to reliably do it from code.  
So what you're doing is probably the best option.

Safihre:
@animetosho Thanks for the feedback :)

animetosho:
There's a compiler check [here](https://github.com/sabnzbd/sabyenc/blob/07291cfe5259d6d6d45bcda22faa0d26b942d954/setup.py#L242). The main reason to be careful is that MSVC uses completely different arguments to GCC-compatible compilers (.e.g no `-march`, `-std` etc), and obviously doesn't target Linux. I also don't think they support anything below ARMv7.

PR #44, opened 2022-04-28
Use a buffer based approach

Safihre:
@animetosho I tried to add a new function that works on a Python-buffer.
However, too my surprise, it's slower than the chunk-method..
This really surprises me.
Can you see something that I am clearly doing wrong?
```
 yEnc C Current  took   312 ms
  yEnc C Buffer  took   312 ms
---
 yEnc C Current  took   328 ms
  yEnc C Buffer  took   437 ms
---
 yEnc C Current  took   281 ms
  yEnc C Buffer  took   421 ms
---
```

### DIFF
--- a/src/sabyenc3.cc
+++ b/src/sabyenc3.cc
@@ -44,6 +44,12 @@ static PyMethodDef sabyenc3_methods[] = {
         METH_O,
         "decode_usenet_chunks(list_of_chunks)"
     },
+	{
+		"decode",
+		decode,
+		METH_VARARGS,
+		"decode(raw_data, raw_data_size)"
+	},
     {
         "encode",
         encode,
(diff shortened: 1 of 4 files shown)


animetosho:
Interesting, you're doing an in-place decode. I'm not sure if that's supported, but I guess it probably works.  
Although you are resizing the buffer at the end, which means it can't be reused afterwards. For the bigger picture, it may be ideal to have a fixed size buffer and reuse it, saving reallocations and GC pressure.

The benchmark here shouldn't really show that much of a difference, as you're essentially just skipping the extra allocation and reducing some of the per-chunk overhead.

As for your question, nothing jumps out to me in the code. I suspect it may be how you're testing it - the [test set](https://github.com/sabnzbd/sabyenc/pull/44/files#diff-102330aa038b83d1fcd47839a7fe986ae8a422bc0decd7faf42a8af0fb8cdc87R27) for the buffer test is rather large (387KB \* 5000 = 1.85GB memory, plus reallocated per test run) whilst the 'current' test is just the same 387KB data tested 5000\*5 times.  
387KB will fit in CPU cache, whilst 1.85GB won't.

Try changing the test by replacing [this line](https://github.com/sabnzbd/sabyenc/pull/44/files#diff-102330aa038b83d1fcd47839a7fe986ae8a422bc0decd7faf42a8af0fb8cdc87R26) with something like:

	list_chunks.append(list(map(lambda x: bytes(bytearray(x)), data_chunks)))

By the way, you might want to be careful with lines [like these](https://github.com/sabnzbd/sabyenc/pull/44/files#diff-507abba6877e4f263d778a243063a187e86588da0259e065cbbd4b8158c02e24R626) - if `cur_char` advances far enough, `cur_char + data_length - MAX_TAIL_BYTES` could point off the end of the buffer, or if `data_length < MAX_TAIL_BYTES`, could point off the front.

Also [this section](https://github.com/sabnzbd/sabyenc/pull/44/files#diff-507abba6877e4f263d778a243063a187e86588da0259e065cbbd4b8158c02e24R608-R611) could probably be simplified to something like:

	 	Py_output_filename = PyUnicode_DecodeLatin1(start_loc, cur_char - start_loc, NULL);

(won't affect performance, just a simplification)

PR #48, opened 2022-05-19
Add VPCLMULQDQ support for CRC calc

animetosho:
Adds the ability to compute CRC using AVX2+[VPCLMULQDQ](https://en.wikipedia.org/wiki/AVX-512#VPCLMULQDQ) feature.

Up to 2x CRC performance on CPUs with the feature (AMD Zen3, Intel Ice Lake or later).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,6 +88,7 @@ class SAByEncBuild(build_ext):
         # Determine compiler flags
         gcc_arm_neon_flags = []
         gcc_arm_crc_flags = []
+        gcc_vpclmulqdq_flags = []
         gcc_macros = []
         if self.compiler.compiler_type == "msvc":
             # LTCG not enabled due to issues seen with code generation where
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -190,6 +190,12 @@ def build_extension(self, ext: Extension):
                 "depends": srcdeps_crc_common,
                 "gcc_x86_flags": ["-mssse3", "-msse4.1", "-mpclmul"],
             },
+            {
+                "sources": ["src/yencode/crc_folding_256.cc"],
+                "depends": srcdeps_crc_common,
+                "gcc_x86_flags": ["-mavx2", "-mvpclmulqdq", "-mpclmul"] if IS_X86 and autoconf_check(self.compiler, flag_check="-mvpclmulqdq") else [],
```
Safihre, line 196: On Windows this is always available if `/arch:AVX2` is supplied?
Because now the  `autoconf_check` is also executed on Windows, but I never made it work there so it's giving a bit of an ugly error: https://github.com/animetosho/sabyenc/runs/6502573699?check_suite_focus=true#step:5:82

Safihre (commented):
Thanks for updating us with the latest and greatest 🥳

setup.py
animetosho: Ah missed that - good catch!  
Should be fixed in latest commit.

Safihre: 👍 Thanks!

PR #49, opened 2022-05-21
Port misc yencode changes for consistency

animetosho:
Not sure how interested you are in this, as it doesn't add much. Rather, it ports changes/fixes made to node-yencode, since initial inclusion, to be more consistent. (the fixes probably mean little to sabyenc)


Safihre:
Thanks! I don't fully understand all changes, but I'm sure they are good.

PR #50, opened 2022-05-21
Enable VBMI2 optimisations

animetosho:
This enables the VBMI2 ISA level already included in yencode. Support for the instruction set is currently limited, but I'm measuring around 15% improvement in decoder performance over AVX2.

This PR also changes the way `simd_detected` works to be more consistent with the actual function selection. Previously it would only check the CPU's capabilities, whereas now it'll also consider the compiler's capabilities.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,6 +89,7 @@ class SAByEncBuild(build_ext):
         gcc_arm_neon_flags = []
         gcc_arm_crc_flags = []
         gcc_vpclmulqdq_flags = []
+        gcc_vbmi2_flags = []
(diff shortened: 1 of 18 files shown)


Safihre:
Cool. My laptop seems to have the `AVX512VL+VBMI2`, but can't reliably measure differences because on Windows the timings jump up and down with every run.

animetosho:
You might be able to get more stable runs if you set your power profile to High Performance, or otherwise lock the CPU frequency.

Thanks for testing+merging!

Issue #76, opened 2023-01-03
Cannot link to OpenSSL on macOS


Safihre:
@animetosho Maybe you know what is going wrong. I thought I could just simply add macOS support, but for some reason it does not seem to search the right folders.
https://github.com/sabnzbd/sabyenc/commit/0146b0ea147515653c903fce923d5c56552de449
Because I located the file as such here, just like on Windows:
<img width="773" alt="Screenshot 2023-01-03 at 14 39 03" src="https://user-images.githubusercontent.com/5703454/210368373-313afca2-4dfb-4cb9-826a-70eff9f300a9.png">

animetosho:
Does macOS use some sort of container/packaging system? If so, maybe it has a private libssl that isn't visible on the system.

Maybe try running the following and using the *libssl* filename printed there.

	otool -L `python3 -c "import ssl; print(ssl._ssl.__file__)"`

Safihre:
Interesting, here it is:
```
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/lib-dynload/_ssl.cpython-39-darwin.so:
	/Library/Frameworks/Python.framework/Versions/3.9/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/Library/Frameworks/Python.framework/Versions/3.9/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
```

animetosho:
Does it work if you use the full path instead of just `libssl.1.1.dylib`?

Can you double check that the function is exported?  
`nm --extern-only /Library/Frameworks/Python.framework/Versions/3.9/lib/libssl.1.1.dylib | grep read`

Also try the following:

	import ctypes
	h=ctypes.CDLL('libssl.1.1.dylib')
	print(h)
	print(h.SSL_read_ex)

(I don't know that much about macOS, so poking in the dark here)

Safihre:
I found a solution by accessing Python runtime information.
Not the most pretty, but it works: 779fdf734b1e4bebc04ac6ed8b5a1e5543042088
Thanks for helping me brainstorm :)

animetosho:
Nice!

Since the full path seems to work, another possibility may be to pass through `ssl._ssl.__file__` to the module - it might work because I think its dependencies get re-exported.

Safihre:
[...]

PR #77, opened 2023-01-06
Find the openssl handle via the python ssl module

mnightingale:
I had a look at making the linking more robust and work for more exotic Python installs such as homebrew.

Related to #76, you said you didn't try this method because the symbols don't get re-exported, I don't really understand how it works but the required dependencies are being loaded and the functions are available.

I've tested and called the method on Linux and Mac (python.org and homebrew).

I also added a test :)

Because it imports `_ssl` itself you don't have to ensure "import ssl" comes first and those lines could move up to also apply to the Windows version.
I think we need the loaded module to stay loaded to keep the handle valid, therefore I don't call `Py_XDECREF(ssl_module)`.

---

Also [SSL_read_ex](https://www.openssl.org/docs/man1.1.1/man3/SSL_read_ex.html) was added in OpenSSL 1.1.1 so if you don't want to do this method then a few of the libraries it tries could be removed since they'll never have SSL_read_ex.

### DIFF
--- a/src/sabyenc3.cc
+++ b/src/sabyenc3.cc
@@ -79,21 +79,17 @@ static bool openssl_init() {
     *(void**)&SSL_read_ex = GetProcAddress(openssl_handle, "SSL_read_ex");
     *(void**)&SSL_get_error = GetProcAddress(openssl_handle, "SSL_get_error");
 #else
- #if defined(__APPLE__)
-    void* openssl_handle = dlopen("libssl.1.1.dylib", RTLD_LAZY | RTLD_NOLOAD);
- #else
-    void* openssl_handle = dlopen("libssl.so", RTLD_LAZY | RTLD_NOLOAD);
-    if(!openssl_handle)
-        openssl_handle = dlopen("libssl.so.1.1", RTLD_LAZY | RTLD_NOLOAD);
-    if(!openssl_handle)
-        openssl_handle = dlopen("libssl.so.3", RTLD_LAZY | RTLD_NOLOAD);
-    if(!openssl_handle)
-        openssl_handle = dlopen("libssl.so.1.0.2", RTLD_LAZY | RTLD_NOLOAD);
-    if(!openssl_handle)
-        openssl_handle = dlopen("libssl.so.1.0.1", RTLD_LAZY | RTLD_NOLOAD);
-    if(!openssl_handle)
-        openssl_handle = dlopen("libssl.so.1.0.0", RTLD_LAZY | RTLD_NOLOAD);
- #endif
+    // Find library at "import ssl; print(ssl._ssl.__file__)"
+    PyObject *ssl_module = PyImport_ImportModule("_ssl");
+    if(!ssl_module) return false;
+
+    PyObject *ssl_module_dict = PyModule_GetDict(ssl_module);
+    if(!ssl_module_dict) return false;
+
+    PyObject *ssl_module_path = PyDict_GetItemString(ssl_module_dict, "__file__");
+    if(!ssl_module_path) return false;
+
+    void* openssl_handle = dlopen(PyUnicode_AsUTF8(ssl_module_path), RTLD_LAZY | RTLD_NOLOAD);
 
     if(!openssl_handle) return false;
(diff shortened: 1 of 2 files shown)


mnightingale:
For the failing Windows tests because of no `import ssl`, how do you want me to resolve them, import ssl in the test or make `openssl_init` always call `PyImport_ImportModule("_ssl")`?

Safihre:
Great! Guess the comment about macOS not exporting the symbols are wrong.
Indeed if we can always import ssl from within the C code, that would be great.

Safihre:
Probably we should not use SSL_read_ex since from what I understand all it does is support >2Gb, which we won't do anyway.
Python 3.8 seems to use the non-ex version, so should we.

mnightingale:
Tests pass :grinning:

A just poked around the 3.8.0 installer and they had the 1.1.1 libraries so just didn't use the newer functions.

I think it's better to stick to doing it however the latest stable Python release does its easier to monitor for any bug fixes.

Most are going to use packaged builds anyway with recent Python.

mnightingale:
I would quite like if we could check it works on ARM, unfortunately I don't have a device to try it on.

src/sabyenc3.cc
```diff
@@ -71,6 +71,9 @@ static struct PyModuleDef sabyenc3_definition = {
 static bool openssl_init() {
     // TODO: consider adding an extra version check to avoid possible future changes to SSL_read_ex
 
+    PyObject *ssl_module = PyImport_ImportModule("_ssl");
```
Safihre, line 74: Officially indeed this should be decref'd, but guess it doesn't matter.

sanderjo:
> I would quite like if we could check it works on ARM, unfortunately I don't have a device to try it on.

I have Ubuntu Linux on ARM. Can I do something for you?

mnightingale:
> I have Ubuntu Linux on ARM. Can I do something for you?

```
$ pip install .
$ python -c "import sabyenc3; print(sabyenc3.openssl_linked);"
```

If it's working it should print "True".

Thanks.

sanderjo:
I used a bit different procedure, but I hope it looks good on my ARM64:

```
sander@nanopineo2:~/git/sabyenc$ uname -a
Linux nanopineo2 5.15.25-sunxi64 #22.02.1 SMP Sun Feb 27 09:24:04 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

```
sander@nanopineo2:~/git/sabyenc$ python3
Python 3.8.0 (default, Dec  9 2021, 17:53:27) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sabyenc3
>>> sabyenc3.__file__
'/home/sander/git/sabyenc/sabyenc3.cpython-38-aarch64-linux-gnu.so'
>>> print(sabyenc3.openssl_linked)
True
>>> 
```

```
>>> dir(sabyenc3)
['SSLWantReadError', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '__version__', 'decode_usenet_chunks', 'encode', 'openssl_linked', 'simd', 'unlocked_ssl_recv_into']
>>> sabyenc3.__version__
'6.0.0'
```

sanderjo:
Also tested on riscv64, and looks good too?

Method:
```
git clone https://github.com/sabnzbd/sabyenc.git openssl_link_sabyenc
cd openssl_link_sabyenc/
git branch -a
git checkout origin/openssl_link
python3 setup.py build
find . | grep -e "\.so$"
cp ./build/lib.linux-riscv64-3.10/sabyenc3.cpython-310-riscv64-linux-gnu.so .
```

```
sander@nezha:~/git/openssl_link_sabyenc$ uname -a
Linux nezha 6.1.0-rc3-d1 #trunk Thu Dec  1 14:53:18 MSK 2022 riscv64 riscv64 riscv64 GNU/Linux

sander@nezha:~/git/openssl_link_sabyenc$ python3 -c "import sabyenc3; print(sabyenc3.__version__); print(sabyenc3.openssl_linked);"
6.0.0
True

```

PR #78, opened 2023-01-09
Refactor openssl_link for stability, safety and better error state handling

mnightingale:
I've split the core code into two methods:

## unlocked_ssl_recv_into_impl
Stays as close to the cpython implementation of `_ssl__SSLSocket_read_impl` as possible to make it simpler to merge in any of their changes.

The main change is including error state hanlding and avoids uses of discarded sockets.

Stips out much of the code that we wouldn't hit:
- Removed option to return bytes, only reads to a buffer
- Decoration of exceptions - I didn't spend too long figuring out what this code does but I think it may be related to the stacktraces and chaining of them
 - You'll get a ssl.SSLWantReadError or generic PyExc_ValueError on failure - the same as previous implementation but hooks the exception from ssl
- Does not support sockets other than non-blocking - the timeout and poll of sock state are not handled

## unlocked_ssl_recv_into
This is the method exposed by the extension, I've changed it to take a `SSLSocket` instead of `PySSLSocket` (`sock` not `sock._sslobj`) - it's neater and the method checks that `_sslobj` exists.

# Miscellaneous
- Moved it into it's own files - easier to manage and seperate the cpython code
- Gets SSLWantReadError from the ssl/_ssl module
- Added a few more macros for SSL_* - they haven't changed in >24 years so I think this is OK - we could get most of them from the ssl module but it's not worth it
- openssl_init() doesn't return anything anymore but added a `bool openssl_linked()` method
- openssl_init() simpler failure handling, hooks a few more things and added missing ref counting

I think that's everything 💥 I'll leave you to add something about license and attributing the cpython code.

### DIFF
--- a/src/unlocked_ssl.cc
+++ b/src/unlocked_ssl.cc
@@ -2,6 +2,90 @@
 
 int (SABYENC_DLL_CALL *SSL_read_ex)(void*, void*, size_t, size_t*);
 int (SABYENC_DLL_CALL *SSL_get_error)(void*, int);
+int (SABYENC_DLL_CALL *SSL_get_shutdown)(void*);
+
+/* Redefined below for Windows debug builds after important #includes */
+#define _PySSL_FIX_ERRNO
+
+#define PySSL_BEGIN_ALLOW_THREADS_S(save) \
+    do { (save) = PyEval_SaveThread(); } while(0)
+#define PySSL_END_ALLOW_THREADS_S(save) \
+    do { PyEval_RestoreThread(save); _PySSL_FIX_ERRNO; } while(0)
+#define PySSL_BEGIN_ALLOW_THREADS { \
+            PyThreadState *_save = NULL;  \
+            PySSL_BEGIN_ALLOW_THREADS_S(_save);
+#define PySSL_END_ALLOW_THREADS PySSL_END_ALLOW_THREADS_S(_save); }
+
+typedef struct {
+    int ssl; /* last seen error from SSL */
+    int c; /* last seen error from libc */
+#ifdef MS_WINDOWS
+    int ws; /* last seen error from winsock */
+#endif
+} _PySSLError;
+
+typedef struct {
+    PyObject_HEAD
+            PyObject *Socket; /* weakref to socket on which we're layered */
+    void *ssl;
+    void *ctx; /* weakref to SSL context */
+    char shutdown_seen_zero;
+    int socket_type;
+    PyObject *owner; /* Python level "owner" passed to servername callback */
+    PyObject *server_hostname;
+    _PySSLError err; /* last seen error from various sources */
+    /* Some SSL callbacks don't have error reporting. Callback wrappers
+     * store exception information on the socket. The handshake, read, write,
+     * and shutdown methods check for chained exceptions.
+     */
+    PyObject *exc_type;
+    PyObject *exc_value;
+    PyObject *exc_tb;
+} PySSLSocket;
+
+static inline _PySSLError _PySSL_errno(int failed, void *ssl, int retcode)
+{
+    _PySSLError err = { 0 };
+    if (failed) {
+#ifdef MS_WINDOWS
+        err.ws = WSAGetLastError();
+        _PySSL_FIX_ERRNO;
+#endif
+        err.c = errno;
+        err.ssl = SSL_get_error(ssl, retcode);
+    }
+    return err;
+}
+
+/* The object holding a socket.  It holds some extra information,
+   like the address family, which is used to decode socket address
+   arguments properly. */
+
+typedef struct {
+    PyObject_HEAD
+            SOCKET_T sock_fd;           /* Socket file descriptor */
+    int sock_family;            /* Address family, e.g., AF_INET */
+    int sock_type;              /* Socket type, e.g., SOCK_STREAM */
+    int sock_proto;             /* Protocol type, usually 0 */
+    PyObject *(*errorhandler)(void); /* Error handler; checks
+                                        errno, returns NULL and
+                                        sets a Python exception */
+    _PyTime_t sock_timeout;     /* Operation timeout in seconds;
+                                        0.0 means non-blocking */
+} PySocketSockObject;
+
+typedef enum {
+    SOCKET_IS_NONBLOCKING,
+    SOCKET_IS_BLOCKING,
+    SOCKET_HAS_TIMED_OUT,
+    SOCKET_HAS_BEEN_CLOSED,
+    SOCKET_TOO_LARGE_FOR_SELECT,
+    SOCKET_OPERATION_OK
+} timeout_state;
+
+/* Get the socket from a PySSLSocket, if it has one */
+#define GET_SOCKET(obj) ((obj)->Socket ? \
+    (PySocketSockObject *) PyWeakref_GetObject((obj)->Socket) : NULL)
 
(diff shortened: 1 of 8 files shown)


Safihre:
Perfect. Thanks so much!

Now if you could add at least a basic set of tests? 🤩
For example a basic read test or something like was added to the cpython PR to test it really reads more than 1 SSL frame: https://github.com/python/cpython/pull/31492
 
You can disable all the yenc tests for now. 😋

mnightingale:
I'm just playing with the tests a little more on a different branch [openssl_link_refactor_tests](https://github.com/mnightingale/sabyenc/tree/openssl_link_refactor_tests)

mnightingale:
I think it's 'done' now, latest test branch run https://github.com/mnightingale/sabyenc/actions/runs/3886895989 (arm still running)

Changes since initial PR:
- Add several tests - shares a ssl echo server with all the tests
- Checks socket argument is an actual SSLSocket instance
- Clearer message when _sslobj cannot be found
- Rejecting blocking sockets
- Fixed a few build warnings - mostly potentially uninitialised pointers, I'm not a C/C++ programmer but got their eventually.

src/unlocked_ssl.cc
```diff
@@ -0,0 +1,254 @@
...
+    do {
+        Py_BEGIN_ALLOW_THREADS;
+        do {
+            retval = SSL_read_ex(self->ssl, mem + count, len, &readbytes);
+            if (retval <= 0) {
+                break;
+            }
+            count += readbytes;
+            len -= readbytes;
+        } while (len > 0);
```
puzzledsab, line 160: I suggested in another branch to add `&& readbytes >= 16384` because in some earlier testing I did it would mean that it was the last block available if there were less. Any opinions on that? I don't know if there are systems that could use smaller max size blocks.

mnightingale:
I've just noticed I forgot to add a test that we actually get  >16 KB responses, I'll take a look at that sometime.
It will need need a few changes to the echo server

src/unlocked_ssl.cc
Safihre: @mnightingale I think this got lost somewhere because I did program. It somewhere, it seems like a good idea since it's the openssl default.

Safihre:
> I've just noticed I forgot to add a test that we actually get >16 KB responses, I'll take a look at that sometime. It will need need a few changes to the echo server

From the test in the cpython PR it seems to be enough to send 64k data, add a `time.sleep(0.5)`  and then see how much you get back. Should be >16K.

Safihre:
@mnightingale I fixed the tests for the yEnc part.

Safihre:
All green! Anything still to add or ready to merge?

src/unlocked_ssl.cc
mnightingale: I did investigate it, but it doesn't fit the logic of the loops.

I believe there are a few TCP congestion protocols going on that result in the first segment being <16KB before quickly ramping up.

- SSL_read_ex in our case is always non-blocking so I would presume it's a very quick call if it has nothing more available
  - returns 1 on success - it read at least one segment
  - returns 0 on failure - may be retryable, call SSL_get_error to find out
- Briefly SSL_ERROR_WANT_READ means we tried to read but it failed because at least 1 segment wasn't yet available

In our python code we before we call unlocked_ssl_recv_into_impl we would poll the socket and know there is data on the underlying socket but there might not be enough to return anything yet.

To illustrate the path and why I believe there isn't anything to gain:

```py
do {
  Py_BEGIN_ALLOW_THREADS;
  do {
      retval = SSL_read_ex(self->ssl, mem + count, len, &readbytes);
      if (retval <= 0) {
          break;
      }
      count += readbytes;
      len -= readbytes;
  } while (len > 0);
  err = _PySSL_errno(retval == 0, self->ssl, retval); // sets err.ssl = SSL_get_error(...)
  Py_END_ALLOW_THREADS;
  self->err = err;

  if (count > 0) {
      break;
  }

  // A load of unused sockstate logic that we mostly don't use except detecting shutdown
  // I kept it so it matches the cpython code/PR
} while (err.ssl == SSL_ERROR_WANT_READ ||
         err.ssl == SSL_ERROR_WANT_WRITE);
```

I'm going to ignore the len calculation because just assume we aren't going to full our buffer so len > 0 is always true.

| iteration | retval | readbytes | count |
| --------- | ------ | --------- | ----- |
| 1         | 1      | 4096      | 4096  |
| 2         | 1      | 16384     | 20480 |
| 3         | 1      | 16384     | 36864 |
| 4         | 0      | 0         | 36864 |

retval <= 0 so break out of the inner loop
count > 0 so break out of the outer loop and return it to python

The only way you'd get past the count > 0 break would be if the first call failed, in which case the error is checked to determine if we should retry or return to python.
It's likely we'll get an SSL_ERROR_WANT_READ since we've already polled and know there is some data on the socket, therefore we'll retry until the first segment is available, and if the next loop after a success fails count will be > 0 and we'll return to python as above.

Basically, to have even gotten to this stage we know there is or will be very shortly some data.
The len > 0 condition is really just making sure we limit our reads to the size of the buffer, as soon as we receive something and there is nothing left to read immediately available we exit.

---

I get what you're saying in that if we've had a 16 KB in the past we could stop reading once we receive a smaller segment, however in my opinion the simplicity of the loop means once the data available is exhausted its going to return anyway.

For the above reasons I don't think there is an issue with reading a large amount of data - I think we've settled on a 1 MB buffer, we are releasing the GIL so other threads can do their work anyway and I think their is a lot of potential with splitting downloading across multiple threads (from my testing very few pools of connections are necessary and it'll depend on total connections - there is no point sharing 8 connections across 4 threads) because multiple downloaders can be reading their data simultaneously.

Quite possibly the maximum read per cycle and hence buffer per connection should be configurable in SAB but that only really helps very slow connections or very low RAM systems - as far as I can tell a whole article is still going to be buffered into memory which the 1MB default helps with, if it starts with a smaller buffer it's going to need to keep growing which is an expensive process.

mnightingale:
Just a test that we receive a response >16 KB I can take a look at adding that this evening.

I'll try and take a look at it this evening but if you want to merge it without that's fine we me I can send another PR.

mnightingale:
Added a test that we get a >16KB response, it's ready to merge now.

Safihre:
Awesome!

Safihre:
@puzzledsab @mnightingale Seems a cpython core developer decided to implement a solution for cpython after all: https://github.com/python/cpython/pull/102214

mnightingale:
It does look a nicer approach and addresses some comments from the other PR, removes the nested do-while loop and supports read and write.

Looks like it has a couple of months to beat the feature freeze for 3.12 so fingers crossed.

Because of the changes it makes I don't think we could port it, so once a release (hopefully) is out with it would you remove the current implementation from sabctools or initially just add a Python version check?

Safihre:
I am highly suspicious if that patch actually solves the problem of reading only 16K at a time
.. But let's see. He didn't add that test.

mnightingale:
Hmm yeah looks that way, "AssertionError: All TLS reads were smaller than 16KB"

mnightingale:
Minor changes required to seem to make it work but I decided against posting it on the cpython issue just in case it causes a delay.

https://gist.github.com/mnightingale/74a8ccca5f6146f79a336b2e3105db17

I think this complicates the flow anyway with more conditions in the main while loop that the current implementations do...while approach wouldn't have.

PR #80, opened 2023-01-16
Add crc32 method utility methods

mnightingale:
It seems simple enough to use the fast CRC calculations from crcutil that use hardware features if available.

I compared no checking, _PyArg_ParseStack and PyArg_ParseTuple and their is barely any difference so went with _PyArg_ParseStack because we get the bounds and type checking.

5M rounds:

```
crc32_combine
7.757969400001457
crc32_combine_parse (this one)
7.880420399997092
crc32_combine_tuple
8.029682699998375
```

Reason for making this a draft PR is I think we may as well try and add methods for the other uses of CRC calculations:

- crc_multiply
- crc_zero_unpad
- crc_2pow

I've not dug into the crcutil interface yet, but I see yencode has a `do_crc32_zeros` method which I think is the same as the unused `crc_zero_pad` so hopefully it shouldn't be too difficult to figure out adding the unpad version.

@Safihre will you want some tests? we can only really test what happens with None arguments or an invalid number of arguments but _PyArg_ParseStack takes care of that, we could test expected CRCs but I'm not sure its necessary being one line that just passes it to crcutil.

@puzzledsab

---

It might be worth thinking of renaming this to sabutils...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -291,6 +291,13 @@ class SAByEncBuild(build_ext):
                 "gcc_flags": ["-Wno-unused-parameter", "-Wno-missing-field-initializers"],
                 'msvc_x86_libraries': ['ws2_32'],
             },
+            {
+                "sources": [
+                    "src/crc32.cc",
+                ],
+                "gcc_flags": ["-Wno-unused-parameter"],
+                "include_dirs": ["src/crcutil-1.0/code", "src/crcutil-1.0/examples"],
+            },
         ]:
             args = {
                 "sources": source_files["sources"],
--- a/src/crc32.cc
+++ b/src/crc32.cc
@@ -42,7 +42,7 @@ PyObject* crc32_zero_unpad(PyObject *self, PyObject *args) {
 }
 
 PyObject* crc32_xpown(PyObject* self, PyObject* arg) {
-    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLongMask(arg);
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLong(arg);
 
     if (PyErr_Occurred()) {
         return NULL;
@@ -50,11 +50,11 @@ PyObject* crc32_xpown(PyObject* self, PyObject* arg) {
 
     crc->XpowN(&n);
 
-    return PyLong_FromUnsignedLongLong(n);
+    return PyLong_FromUnsignedLong(n);
 }
 
 PyObject* crc32_xpow8n(PyObject* self, PyObject* arg) {
-    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLongMask(arg);
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLong(arg);
 
     if (PyErr_Occurred()) {
         return NULL;
@@ -62,5 +62,5 @@ PyObject* crc32_xpow8n(PyObject* self, PyObject* arg) {
 
     crc->Xpow8N(&n);
 
-    return PyLong_FromUnsignedLongLong(n);
+    return PyLong_FromUnsignedLong(n);
 }
--- a/src/crc32.cc
+++ b/src/crc32.cc
@@ -42,7 +42,7 @@ PyObject* crc32_zero_unpad(PyObject *self, PyObject *args) {
 }
 
 PyObject* crc32_xpown(PyObject* self, PyObject* arg) {
-    crcutil_interface::UINT64 n = PyLong_AsUnsignedLong(arg);
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLong(arg) & 0xffffffff;
 
     if (PyErr_Occurred()) {
         return NULL;
@@ -54,7 +54,7 @@ PyObject* crc32_xpown(PyObject* self, PyObject* arg) {
 }
 
 PyObject* crc32_xpow8n(PyObject* self, PyObject* arg) {
-    crcutil_interface::UINT64 n = PyLong_AsUnsignedLong(arg);
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLong(arg) & 0xffffffff;
 
     if (PyErr_Occurred()) {
         return NULL;
--- a/src/crc32.cc
+++ b/src/crc32.cc
@@ -1,0 +1,69 @@
+#include "crc32.h"
+#include "crcutil-1.0/examples/interface.h"
+
+extern crcutil_interface::CRC *crc;
+
+PyObject* crc32_combine(PyObject *self, PyObject *args) {
+    uint32_t crc1, crc2;
+    size_t length;
+
+    if(!PyArg_ParseTuple(args, "IIn:crc32_combine", &crc1, &crc2, &length)) {
+        return NULL;
+    }
+
+    crcutil_interface::UINT64 crc1_ = crc1, crc2_ = crc2;
+    crc->Concatenate(crc2_, 0, length, &crc1_);
+
+    return PyLong_FromUnsignedLong((uint32_t) crc1_);
+}
+
+PyObject* crc32_multiply(PyObject *self, PyObject *args) {
+    uint32_t crc1, crc2;
+
+    if(!PyArg_ParseTuple(args, "II:crc32_multiply", &crc1, &crc2)) {
+        return NULL;
+    }
+
+    crcutil_interface::UINT64 crc1_ = crc1, crc2_ = crc2;
+    crc->Multiply(crc2_, &crc1_);
+
+    return PyLong_FromUnsignedLong((uint32_t) crc1_);
+}
+
+PyObject* crc32_zero_unpad(PyObject *self, PyObject *args) {
+    uint32_t crc1;
+    size_t length;
+
+    if(!PyArg_ParseTuple(args, "In:crc32_zero_unpad", &crc1, &length)) {
+        return NULL;
+    }
+
+    crcutil_interface::UINT64 crc_ = crc1;
+    crc->ZeroUnpad(length, &crc_);
+
+    return PyLong_FromUnsignedLong((uint32_t) crc_);
+}
+
+PyObject* crc32_XpowN(PyObject* self, PyObject* arg) {
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLong(arg);
+
+    if (PyErr_Occurred()) {
+        return NULL;
+    }
+
+    crc->XpowN(&n);
+
+    return PyLong_FromUnsignedLongLong(n);
+}
+
+PyObject* crc32_Xpow8N(PyObject* self, PyObject* arg) {
+    crcutil_interface::UINT64 n = PyLong_AsUnsignedLongLong(arg);
+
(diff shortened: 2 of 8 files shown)


Safihre:
Should we be using `_` methods? These are specially designated as internal functions by cpython.
Especially since the difference so extremely small compared to parse tuple?
We can probably add some simple tests for correctness of the calculation?

mnightingale:
I'll switch to parse tuple, I'm not really sure why parse stack is private api is goes hand in hand with fastcall.
Maybe it's because fastcall was private api until 3.10.
As you say there is barely any difference and we aren't going to be calling it 5M times!

I'll add some simple tests and try and figure out adding the other calculations this evening.

mnightingale:
I switched to ParseTuple earlier and I've figured out how to expose a multiply method but I don't like my current approach to adding it, it's about 10x faster on my oldish Mac.

It's using yencodes crc instance via `extern crcutil_interface::CRC *crc;` but I had to add a Multiply method to `interface.h/cc`.

It looks like interface.h/cc was supposed to be an example interface and it instructs you to make your own.
Therefore I would prefer to leave the example unchanged and add our own interface to sabyenc and create an instance of it separate from yencode.

@puzzledsab This is assuming we do need replacements for all of crc_multiply, crc_zero_unpad and crc_2pow?

Perhaps @animetosho has an opinion on my suggested approach?

mnightingale:
Branch with a messy implementation of multiply and unpad on [feature/crc32_methods_temp](https://github.com/mnightingale/sabyenc/tree/feature/crc32_methods_temp).

2pow is simple enough but I think I'll expose what crcutil calls Xpow8N, which is just XpowN(n << 3) AKA n*8 since all uses of it are timesing by 8 the input.

animetosho:
> since all uses of it are timesing by 8 the input.

Unfortunately not the case with unpadding - the \*8 needs to occur *before* bit-flipping, not [after](https://github.com/mnightingale/sabyenc/commit/7907bbcd16af76df2cf77c7add10f772dfd34d0c#diff-7b6c482844507935bbec743f89069a59254e06e52ff188ec27d244a8bafdc81dR174).

	[inverting 1]
	~(1 * 8) = 0xFFFFFFF7  (correct)
	 ~1 * 8  = 0xFFFFFFF0  (incorrect)

Unpadding isn't strictly required, as you could instead pad the CRC computed from decoded articles, to match the padding on PAR2's CRC. Though unpadding PAR2's CRC does have the nice effect of giving you the file CRC32, should that ever be useful.

> Perhaps @animetosho has an opinion on my suggested approach?

Exposing equivalents to the Python functions seems to be the most straightforward. You may also want to consider things at a higher level, for example, if you've got a list of CRCs and sizes, perhaps pass those in instead of doing the loop in Python.

You could probably also just do full, uncached computation if you're doing it all in C. I suggested the caching techniques as I suspected that Python wouldn't be fast enough otherwise.

On the other hand, if a goal is chasing as much performance as possible, there's more efficient ways to write some functions, for example, multiplying without a loop:

```c
uint32_t crc_multiply(uint32_t a, uint32_t b) {
	// do the actual multiply
	__m128i prod = _mm_clmulepi64_si128(_mm_cvtsi32_si128(a), _mm_cvtsi32_si128(b), 0);
	
	// prepare product for reduction
	prod = _mm_add_epi64(prod, prod); // bit alignment fix, due to CRC32 being bit-reversal
	prod = _mm_slli_si128(prod, 4);   // straddle low/high halves across 64-bit boundary - this provides automatic truncation during reduction
	
	// do Barrett reduction back into 32-bit field
	const __m128i reduction_const = _mm_set_epi32(
		1, 0xdb710640, // polynomial * 2
		0, 0xf7011641  // 2**63 / polynomial
	);
	__m128i t = _mm_clmulepi64_si128(prod, reduction_const, 0);
	t = _mm_clmulepi64_si128(t, reduction_const, 0x10);
	t = _mm_xor_si128(t, prod);
	
	return _mm_extract_epi32(t, 2);
}
```

This does require an x86 CPU with the CLMUL instruction, which means you'd have to implement a CPU feature check dispatcher. I think it chasing this may be more work than actual benefit, but it's certainly something that could be done.

mnightingale:
I only have basic understanding of what's going on but crcutil defines:

```c
// returns ((x ** n) mod P).
Crc XpowN(uint64 n) const {
  Crc one = this->one_;
  Crc result = one;

  for (size_t i = 0; n != 0; ++i, n >>= 1) {
    if (n & 1) {
      result = Multiply(result, this->x_pow_2n_[i]);
    }
  }

  return result;
}

// Returns (x ** (8 * n) mod P).
Crc Xpow8N(uint64 n) const {
  return XpowN(n << 3);
}
```

This is what I have currently as a ZeroUnpad method, I've ran it millions of times and it matches the python result.
Is it not the same calculation? is there some value of crc/length that it will fail for?

```c
virtual void ZeroUnpad(UINT64 bytes,
                       /* INOUT */ UINT64 *lo) const {
  SetValue(crc_.Base().Multiply(GetValue(lo, NULL) ^ crc_.Base().Canonize(),
                                crc_.Base().Xpow8N(bytes ^ 0xFFFFFFFF)) ^ crc_.Base().Canonize(),
           lo,
           NULL);
}
```

For zeros = 1:

```
sabyenc Xpow8N(1 ^ 0xFFFFFFFF) 1701301141

// inverse = ((zeroes * 8) % 0xFFFFFFFF) ^ 0xFFFFFFFF
python crc_2pow(inverse) 1701301141
```

I do now realise that the python method crc_2pow (x ** n) isn't the same as XpowN ((x ** n) % P) and since that is used in other calculations I'll need to figure out how to expose it - or just leave that one in Python.

I'm afraid the rest of you post is pretty beyond me; however I don't think chasing even more performance is necessary for our uses, realistically it's my understanding we are going to run crc on something like 20GB/2MB = ~10,000 * (how many operations we do per article) operations and that's over the course of a whole download so the actual impact of any optimisation is pretty minimal - I'm just keen to reduce SABs CPU usage and this seemed like an easy win since we already use a library for fast crc calculations and this should be a simple drop in to replace crc32calc.py

animetosho:
Yeah you're right, sorry, and thanks for confirming.  
I had assumed calculations as 32-bit ops, but when done in 64-bit, it works.

Whilst looking into this, I also discovered that `(X << k) % ~0` is equivalent to `X <<< k` (where <<< denotes a bit-rotate) - trick might come in handy some day 😆

> I do now realise that the python method crc_2pow (x ** n) isn't the same as XpowN ((x ** n) % P) 

It's the same - all operations in a GF are modulo the polynomial, so I didn't bother putting the '% P' in the comment.

Safihre:
@mnightingale I will wait for you to mark the PR as ready for review, since the content docussion is definitely beyond me 😂

mnightingale:
:rofl: I'm almost done, all the methods are implemented, no quite ready to commit but mostly just have some tests to write.

@animetosho what argument type should be allowed for the XpowN/Xpow8N method, what we have in Python [crc32calc.py](https://github.com/sabnzbd/sabnzbd/blob/ac32d3c835dfc6ef557e4236ed56d8f82fe6c1bc/sabnzbd/utils/crc32calc.py#L46-L55) says it takes an int but Python seems to let you pass (2**32)*8 and it'll output 8388608.

I presumed 2**32-1 (uint_max) should be the maximum input allowed so have used PyLong_AsUnsignedLong (returns unsigned long or throws an exception on overflow) but then it doesn't work on the same range of inputs as the Python implementation.

Calculations using crcutil appear to be on an unsigned long long, so I could change to PyLong_AsUnsignedLongLong and get the same results as Python, but I just wanted to check we should be allowing such large values?

animetosho:
I generally recommend 64-bit ints for anything that deals with byte lengths, especially since crcutil takes them, but I doubt you'll exceed the range of a 32-bit int any time soon.  
So `PyLong_AsUnsignedLongLong` would be my recommendation.

Safihre:
Nice! Now.. Tests? 😇
We could compare with python implementation.

Just to be sure: is any of the calculations heavy enough to make us lift the GIL? No right?

mnightingale:
A couple of minor things to change but mostly just tests to write.

The calculations are extemely fast - on balance they are already fast enough with the python implementation but most of what we needed to make them more efficient was already here and only really a few minor changes to expose the other calculations.
Since they are so fast it's not worth lifting the GIL, I'd expect it would actually be slower.

100K rounds on my Macbook:

```
python
concat 5.660315043001901
multiply 1.127905314002419
zero_unpad 7.612788589991396
2pow 6.468656869983533

sabyenc
concat 0.23121889101457782
multiply 0.13985193500411697
zero_unpad 0.21951808899757452
2pow 0.1605083069880493
```

Safihre:
👍

mnightingale:
I've made the tests by comparing zlib were possible and the python versions [crc32calc.py](https://github.com/sabnzbd/sabnzbd/blob/feature/buffered_reader/sabnzbd/utils/crc32calc.py)

Before I make this ready for review when you get a chance @animetosho would you mind checking my crc32_xpown/crc32_xpow8n.
I noticed there are cases where the Python implementations (crc_2pow and crc_2pow_slow) don't agree with each other so I'm not certain I'm getting the correct value - although it always seems to agree with one of them

Using script https://gist.github.com/mnightingale/03a8e7d42b09d0a01cc02bb9973114a3 I get the following output:

> 8589934590
> crc_2pow 2147483648
> **crc_2pow_slow 3681617473**
> crc32_xpown 2147483648 
> 
> 112233445566
> crc_2pow 2636599254
> **crc_2pow_slow 1289198061**
> crc32_xpown 2636599254 
> 
> 18446744073709551615 Xpow8N
> **crc_2pow 2147483648**
> crc_2pow_slow 3742066410
> crc32_xpown 3742066410 

It's possibly because I'm testing the extremes of allowed Python values that will never usually be hit but I just don't know which is the correct value.

Thanks

animetosho:
> Possibly it's only for values >4294967295 ?

Yeah. `crc_2pow_slow` wasn't written to deal with values exceeding the 32-bit range.  
`crc_2pow` should work correctly with any size. `Xpow8N` will likely fail for 2^61 or larger, because you start losing bits when the `<<3` is applied.

mnightingale:
Right I'm happy with it now, I've tested it against the limits of `0 - 2^32-1` and matched the Python implementation.
I've done `n % 0xffffffff` as suggested, I guess I could have left it out and required that the caller does it, but it really makes no difference.

We are never going to get close to passing `2^32-1` anyway since I think the largest value we pass in is the article size (we could have about 130 MB articles before the `% 0xffffffff` has any effect.

Safihre:
Thanks a lot!

Safihre:
@mnightingale Stupid question maybe, but when do we need `crc32_xpow8n`? In SABnzbd we only use `crc32_xpown`, right?

mnightingale:
It's just a convenience method that I noticed crcutil has, it applies *8 to the input by bit shifting instead of multiplication.
It's not strictly required but `crc_xpown(nzf.nzo.article_size * 8)` is the same as `crc_xpow8n(nzf.nzo.article_size)`

PR #81, opened 2023-01-21
From a bytearray, acquire a buffer, decode and resize the bytearray

mnightingale:
Related to the discussion on #79

Should perhaps rename the method decode or decode_bytearray ?

The general flow is:
- Pass in bytearray
- Get a buffer and its size
- Release GIL
- do_decode
- Acquire GIL
- Release buffer
- Resize bytearray

It uses PyByteArray_Resize to resize the bytearray therefore we need to receive an argument of type bytearray (so it's read/write), acquire a buffer and release it before resizing.

PyByteArray_Resize only reallocates is the new size is <50% than currently allocated which is a bit smarter than the one for bytes - hence the custom method was required.

Possibly in the future if the download+decode process is simplified the bytearray wouldn't be required and instead the decoded size could be returned or decode into a new buffer. (memoryview for download => decode into memoryview/newbuffer => copy into article cache).

### DIFF
--- a/src/sabyenc3.cc
+++ b/src/sabyenc3.cc
@@ -47,8 +47,8 @@ static PyMethodDef sabyenc3_methods[] = {
     {
         "decode_buffer",
         decode_buffer,
-        METH_VARARGS,
-        "decode_buffer(raw_data, raw_data_size)"
+        METH_O,
+        "decode_buffer(raw_data)"
     },
     {
         "encode",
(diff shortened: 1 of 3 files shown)


mnightingale:
Patch for sab https://gist.github.com/mnightingale/c51b172ee0b2196862f79b101f082665 (I missed a bit on the previous one)

Safihre:
Great!
Did you manage to test it with the SABnzbd patch to see how it performs?

mnightingale:
I've ran it with the above patch, as expected it should perform the same.
The only difference should be that it's less risky because it removes possible crashes due to concurrent access.

I didn't use PyArg_ParseTuple because it's now only one argument, but If I had it would have been "Y" instead of "S" so that we'd get the PyByteArrayObject.

The only real difference is we then properly request a buffer which is guaranteed to be continuous - I don't think this is guaranteed with ob_sval+length and could therefore have caused crashes due to access to unowned memory.

Because we are getting a buffer that locks the object it must be released before it can be resized, for simplicity of code flow I handle this after the "finish" label which means I need to initialise output_len with the original length in case we reached that point from an error condition - it wont do anything in error cases anyway because "original length" == "new length" and it early returns.

The standard PyByteArray_Resize is also safer than the custom method needed for bytes because it skips a bunch of checks.

Safihre:
@mnightingale thanks again!
If you are interested in receiving a financial gift for your contributions to SABnzbd, contact me at safihre@sabnzbd.org
It's something I offer to all contributors that help implement major changes or smaller sustained changes 😊 We do have some ad income and I love to share that with people that take a real interest in helping!

PR #85, opened 2023-02-26
Add reading of yenc start and end

Safihre:
@puzzledsab I extended `sabctools` to also output start/end of an article, so we can proceed with using sparse files efficiently.
Note that start/end can also be 0, if one article contains all data of a single file.
I could adjust it that it then sets `end` to the `size`, if that's useful? 

CC: @mnightingale


mnightingale:
@puzzledsab might be able to advise better since they have looked at this recently but I think we might want some logic to deal with yenc starting indexes at 1?

i.e. I think `begin=1 end=100` would really mean seek(0) and write data of length 100

> Note that start/end can also be 0, if one article contains all data of a single file.
I could adjust it that it then sets end to the size, if that's useful?

I'm not sure if we even need the yenc the based length?
I recall there was some discussion about keeping partial articles in which case the yenc based length may not match the actual length of the data - so we'd need to use len(data) anyway?

However, possibly we'd want to pass back if yenc[end] == bytearray length so we know we can do sequential writes and avoid unnecessary seeks - it depends if seek to the current offset is no-op.

puzzledsab:
The useful data are the index of the beginning (begin - 1) and the length (end - begin + 1) so it's easier if those are pre-calculated. If it's a single part then that's useful to know, for instance by setting the length at 0, because it means that there's no point in setting sparse mode or chopping of the end of an article that is longer than it should be. For other parts we need the length to know if sparse mode needs to be enabled (if it's too short) and make sure that articles that are too long don't try to overwrite existing data. If it's too short then it can be padded instead of enabling sparse mode but the yenc length is still needed.

I haven't looked at this for a while. I need to update the branch with the current develop and check if what I have makes sense. There are some parts that I couldn't understand why are the way they are when I looked at it now.

Safihre:
I don't know if we need to really take care of the case when data has invalid positions. I think nzbget also always assumes the values are correct and has no fallback to sequential writes. Maybe you have some insight into that @mnightingale? You're better with C++ than me 😄

mnightingale:
Just for simplicity these are the variables nzbget uses when parsing the headers, it's the same as @puzzledsab explained above.

[brackets] = from a yEnc header line

```
articleFileSize = [size]
articleOffset = [begin] - 1
articleSize = [end] - [begin] + 1
```

As you expected nzbget does just trust the header values, however it does have a couple of safeguards `articleSize <= 0 || articleSize > 1024*1024*1024` but I think the second is mostly because it uses articleSize to allocate memory for the decode so it's just avoiding someone constructing malicious articles but isn't a concern for us.

> Note that start/end can also be 0, if one article contains all data of a single file.
I could adjust it that it then sets end to the size, if that's useful?

That's how nzbget handles it, but I'm not sure if we actually need all 3?
I think we just need `articleFileSize` and `articleOffset` because that would be enough to create a sparse file when required and seek to the correct offset then we'll just write whatever we decoded.

But yes I agree I don't think we need to do any checking of invalid offsets or non-matching sizes.

puzzledsab:
I'm not sure if I understand. Do you want to always use sparse mode and seek before every write? That's slower in my tests than writing sequentially like now. If the part is too long then it can overwrite existing correct data. It may also cause a crash. In my tests it works but I saw someone having problems with not being able to overwrite existing data in sparse files on Stackoverflow or something. It'll probably very rarely be too big, though.

If we make sure all the parts are in the correct spot then it would be easier to implement a retry for only bad or missing parts.

mnightingale:
I wasn't really worrying about how we'd use them just what we'd possibly need, I guess it can't hurt to have all 3 variables.

What I was thinking is that if we are going to trust the headers then I didn't think we'd need the expected size of the part [begin] to [end] but if you think malformed parts happen and we should potentially ignore the decoded length then maybe we do but then why trust the headers over the actual data length?

I wouldn't expect to seek for every article, maybe something like keeping track of the current position and seeking only when required, so I'd still expect sequential writing without any seeking most of the time.

Some areas I'd like to address are:
1. Be able to use a fairly small cache (somewhere around 200)
2. Correct offsets so articles can be retired, if a new server is added, retention tweaked, etc.
3. Correct offsets so files can be repaired via exotic methods (P2P) or combining multiple posts
4. Possibly keep a reference to the open file rather than openining it on every assemble

Safihre:
Yes I was thinking the exact same implementation!

Safihre:
I started on an implementation and so far it seems to work well, indeed with a lot less safeguards than @puzzledsab, more like NZBGet.
Will present PR later.

Safihre:
Closing in favor of #89

PR #87, opened 2023-03-07
fix inconsistent license info

### LINKED ISSUE #86: inconsistent license info
All src/*.{cc,h} files carry GPL-2+ headers, but setup.py and LICENSE.md say LGPLv3.

jcfp:
closes #86


src/yenc.cc
```diff
@@ -1,5 +1,6 @@
 /*
- * Copyright 2007-2023 The SABnzbd-Team <team@sabnzbd.org>
+ * Copyright 2003, 2011 Alessandro Duca <alessandro.duca@gmail.com>
```
Safihre, line 2: All code from the original `yenc` is no longer present, so we can remove this and any references to Alessandro (I think).

Safihre:
Thanks! 👍

Safihre:
✅

PR #92, opened 2023-05-26
Extract size attribute from ybegin line

mnightingale:
We're going to need this for allocating sparse files - I'll hopefully submit a PR sometime over the weekend.

I also noticed a use of part_size perhaps to explore, instead of copying the connection buffer `nw.get_data_buffer()` it's probably quicker and a bit more memory efficient to allocate an empty bytes of length part_size, read from nw.data_view and decode into the new bytes.


Safihre:
Looking at this, do we still need the article size? Could just use len(data)

mnightingale:
It's probably unlikely we'll ever use article size in Python, but there is a possible optimisation which could be made in the future by using it in the C extension for decoding.

Currently:
- Download into nw.data
- Malloc and memcpy with `nw.get_data_buffer()` - whole yenc encoded data so always larger than the decoded version
- PyByteArray_Resize to the decoded length but this is cheap anyway since if I recall it only releases memory if the size reduction is significant.

Possible:
- Download into nw.data
- Pass memoryview to `yenc_decode`
- `yenc_decode` can use the article size to allocate a new bytes or bytearray of the known decoded length
- `do_decode` reading from the memoryview/buffer and writing to the new correctly sized output buffer which can be returned to Python

The advantage would be smaller memory allocation and remove the memcpy.

mnightingale:
What's the best way to use this in a sparse PR?

Are you able to publish a release?

Safihre:
Done

animetosho:
> `do_decode` reading from the memoryview/buffer and writing to the new correctly sized output buffer which can be returned to Python

This issue with this is that you have to *trust* the given size being correct.

PR #98, opened 2023-08-14
Fix parsing 64 bit representations of crc32

### LINKED ISSUE #96: Incorrect CRC32 failure
@animetosho @mnightingale sorry to bother you again for help, but I am not sure how to fix this.

A user reported SABnzbd failing an NZB due to article CRC errors, while another downloader succeeded.
When I disable the article CRC32 handling in SABnzbd, indeed the download succeeds and verifies using par2, so `sabctools` seems wrong.
I _think_ it has to do with the formatting of the CRC32:
```
=yend size=409600 part=1 pcrc32=ffffffff79b5066a
```
If I use the Python verification of the decoded content it produces `79b5066a`, so it matches, minus the `ffffffff`.

I added a test-case in this branch:
https://github.com/sabnzbd/sabctools/commit/a2eddde4c19f7dcde1c36439820d4f9a94d8b565

How do we handle that in `sabctools`?

mnightingale:
Fixes #96 

After a proper look at this I think it is only broken on Windows with ﻿﻿strtoul overflowing and returning ULONG_MAX, so I've replaced it with strtoull and since we assign it to a uint32_t we should be discarding the upper 32 bits.

@animetosho do you see any problems with this? I'm not clear why it worked on Linux/Mac but not Windows as far as I can see they should have the same overflow behaviour?


Safihre:
It's that simple of a fix? Nice 😎

animetosho:
> I'm not clear why it worked on Linux/Mac but not Windows as far as I can see they should have the same overflow behaviour?

It's probably because the MSVC compiler considers [the `long` type to be 32 bits](https://learn.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=msvc-170), whilst I recall GCC/Clang x86-64 considers it to be 64 bits (I'm guessing it's 32-bit on x86-32).

I think `long long` is generally 64 bits, though the C standard makes no guarantee of it. It might be good enough to not really matter though.

sanderjo:
@animetosho should we have a comment in the (cc?) source code explaining that, so that later on we still understand it?

Safihre:
Or (if I understand correctly) should we change the date type to a 64bit one? Or would that be invalid?

animetosho:
> should we have a comment in the (cc?) source code explaining that, so that later on we still understand it?

My personal opinion is that it makes sense to explain why 64 bits is being read for a CRC32, but what should/shouldn't be commented on is subjective.

> Or (if I understand correctly) should we change the date type to a 64bit one?

I don't believe the C standard has specific width versions of `strtoull` and the like, unfortunately. (MSVC does have [`_strtoui64`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strtoui64-wcstoui64-strtoui64-l-wcstoui64-l?view=msvc-170), but it's non-standard)

If you strictly want 64 bit, it may be best to implement the conversion oneself.

mnightingale:
I think it's probably worth adding a comment that we have used `strtoull` because to anyone unfamilar with this issue it is seeming the wrong method for CRC32, it could be useful if the code is ever referred to for porting to another language (someday I'll try and persuade @Safihre to rewrite in Go).

I'm a little confused about the rest, we want a `uint32_t` because it's supposed to be a crc32, so if we receive more bits than fit we want to discard them, so as far as I can see the PR has the correct types without further changes?

The tests pass on 64 bit Windows, do we need to test 32 bit? I've tested 32 bit Linux.

Safihre:
CI runs automatically on 32bit because we make a 32bit SABnzbd release 😊
https://github.com/sabnzbd/sabctools/actions/runs/5859415929/job/15885360940?pr=98

Safihre:
And yes, let's add a comment then! 👍

PR #100, opened 2023-09-12
Fix OpenSSL 3 linking on Windows

mnightingale:
Python 3.11.5 and 3.12 RC Windows distribution come with OpenSSL 3.0.x, so updated search to try the 3 DLL and fallback to trying 1.1

Related https://github.com/sabnzbd/sabnzbd/issues/2421#issuecomment-1715221987


Safihre:
Of course! Thanks!

PR #102, opened 2023-09-20
Add bytearray_malloc

Safihre:
@puzzledsab @mnightingale 
Hoping on your input on this one :)

I was checking the output of the `yappi` profiler [after skipping all the time spent during imports and startup](https://github.com/sabnzbd/sabnzbd/tree/feature/yappi) on SABnzbd. 
I checking not only calls that are high on the list for `Own time`, but also calls that take long even if called only a few times.
One that stood out is `NewsWrapper.get_data_buffer` in which we take a copy of the connection's buffer to pass it along to the `yenc_decode`. Back when making this we considered creating a new buffer for each article, but I think it was @puzzledsab's testing that revealed this was slower. 
Just to be sure I created a `timeit` script (see below) that compares new-buffer-each-time vs copy-buffer-contents. It seems new-buffer is a little faster, but not much. I investigated a bit more and found that cpython does [a `memset` to 0 when creating a new `bytearray`](https://github.com/python/cpython/blob/754519a9f8c2bb06d85ff9b3e9fe6f967ac46d5c/Objects/bytearrayobject.c#L821) which is slow of course.
However, the C-API function `PyByteArray_FromStringAndSize` does not, it [does a `malloc`](https://github.com/python/cpython/blob/754519a9f8c2bb06d85ff9b3e9fe6f967ac46d5c/Objects/bytearrayobject.c#L135).
Since we do not care about the contents of the buffer, it seems to me that a garbage buffer is fine? So I created a utility function `bytearray_malloc` in `sabctools`. As expected, it's magnitudes faster.
```
Copy (current): 0.08248800004366785
New buffer:     0.06513279990758747
Malloc buffer:  0.0016288000624626875
```
What do you think of this change? Both conceptually and the actual implementation? I can move it to it's own `.c` file and add tests.

<details><summary>Script</summary>

```python
import random
import string
import timeit
import sabctools

length = 800_000
result_str = (''.join(random.choice(string.ascii_lowercase) for i in range(length))).encode()
start = bytearray(result_str)
memv = memoryview(start)


def current(memv_in):
    return bytearray(memv_in[:length-500])


def default(byt):
    finish = byt
    temp = bytearray(800_000)
    new_memv = memoryview(temp)
    return finish


def new(byt):
    finish = byt
    temp = sabctools.bytearray_malloc(800_000)
    new_memv = memoryview(temp)
    return finish


for _ in range(10):
    print("---")
    print("Copy (current):", min(timeit.repeat(stmt="current(memv)", setup="from __main__ import current, default, new, start, memv", number=1000)))
    print("New buffer:    ", min(timeit.repeat(stmt="default(start)", setup="from __main__ import current, default, new, start, memv", number=1000)))
    print("Malloc buffer: ", min(timeit.repeat(stmt="new(start)", setup="from __main__ import current, default, new, start, memv", number=1000)))
```

</details>


Safihre:
Thanks for the feedback! 💯

Safihre:
@mnightingale If you have the time, is it maybe clear to you why all tests succeed for sabctools but my implementation in SABnzbd fails?
https://github.com/sabnzbd/sabnzbd/tree/feature/new_sabctools

It must be related to how I pass the buffer to the decode function, but I'm just not seeing it..

mnightingale:
I've not had chance to look at why your branch fails, but at a glace there seems to be a lot of changes and I'm not really sure why.

I was thinking something like this:

```
# Maybe there is a nicer way of doing len(self.data or [])
self.data = sabctools.bytearray_malloc(max(NNTP_BUFFER_SIZE, len(self.data or [])))
self.data_view = memoryview(self.data)

...

# Then in decode slice the memoryview since this won't make a copy
sabnzbd.decoder.decode(article, nw.data_view[: nw.data_position])
```

In sabctools, revert the decoder changes and instead of requiring a bytearray with PyByteArray_Check use PyObject_CheckBuffer to let anything supporting the buffer protocol be used.
Due to slicing of the memoryview we would not need the size.
Remember to remove PyByteArray_Resize or type check it.

The tricky bit is what to return, maybe a new memoryview slice but I haven't dug into the C api calls to see how we'd do that, you could test just adding the length to the returned tuple and saves the cost if any of creating a new memoryview.

There would be a bunch of changes to make everything expect a memoryview, for uu probably just make a copy for now.

increase_data_buffer should probably increase by a smaller amount (for instance 50% of itself or NNTP_BUFFER_SIZE) since the whole buffer will remain in memory and I'd suggest something like max(NNTP_BUFFER_SIZE, existing_size) so we presume the previously used size will be required again to avoid repeating a create + resize flow.

mnightingale:
Just taking a step back are you sure you are looking at the correct path in bytearrayobject.c won't we currently be doing [/* Use the buffer API */](https://github.com/python/cpython/blob/754519a9f8c2bb06d85ff9b3e9fe6f967ac46d5c/Objects/bytearrayobject.c#L828)

---

> I was just faced with the need to know the size before we can decode, so that's why I'm passing around the length.

Passing in a slice of the memoryview saves needing to pass around the length.

> In the end we do need to end up with a resized bytearray to write to disk. So somewhere we need a resize, right?

A disadvantage of switching around when a new buffer is created is that it is a bit more wasteful with memory, hence my suggestion of reducing how much increase_data_buffer grows by.

A logical advantage of your way and the way it works currently is we only keep the used part of the buffer via the copy and resize the array after decoding, however yenc is very efficient so the size doesn't reduce much but more importantly if I recall PyByteArray_Resize doesn't always actually do any resizing unless the unused capacity is significant.

Passing memoryview slices around will keep the whole backing buffer alive even if we subslice it after decoding.

Certainly memoryviews could go in the article cache and eventually be written to disk, but we're keeping the whole buffer around until it's written and disposed of whereas currently we are probably just wasting the space requried for headers and yenc overhead.

---

Another approach which takes advantage of your findings with malloc and because we now decode inline could be if yenc_decode returned a new bytearray.
- read from a slice of data_view
- find encoded yenc data (we already have this) and malloc a new bytearray
- decode into the new array
- resize to decoded length
- return bytearray

Safihre:
Yes, those steps sound like a much better plan. Indeed I forgot that the decode is inline so we can do it this way!

PR #103, opened 2023-09-24
Use memorview as input and a new bytesarray as output

Safihre:
@mnightingale Like this? :)


Safihre:
Indeed the changes needed in SABnzbd are small: https://github.com/sabnzbd/sabnzbd/tree/feature/new_sabctools
(will fix uu later)

src/yenc.cc
```diff
@@ -204,6 +202,10 @@ PyObject* yenc_decode(PyObject* self, PyObject* Py_bytesarray_obj) {
         goto finish;
     }
 
+    // Create our destination bytearray
+    Py_output_bytearray = PyByteArray_FromStringAndSize(NULL, Py_buffer_obj->len);
```
mnightingale, line 206:
```suggestion
    Py_output_bytearray = PyByteArray_FromStringAndSize(NULL, yenc_data_length);
```

Should probably also guard against 0 length if that is possible by checking if PyByteArray_FromStringAndSize returns NULL

tests/test_decoder.py
```diff
@@ -75,15 +75,15 @@ def test_ref_counts():
     data_plain = read_plain_yenc_file("test_regular.yenc")
     data_out, filename, filesize, begin, end, crc_correct = sabctools_yenc_wrapper(data_plain)
     # data_plain and data_out point to the same data!
```
mnightingale, line 77:
```suggestion

```

src/yenc.cc
```diff
@@ -78,26 +79,23 @@ PyObject* yenc_decode(PyObject* self, PyObject* Py_bytesarray_obj) {
...
-    if (!PyByteArray_Check(Py_bytesarray_obj)) {
-        PyErr_SetString(PyExc_TypeError, "Expected bytearray");
+    if (!PyMemoryView_Check(Py_memoryview_obj)) {
+        PyErr_SetString(PyExc_TypeError, "Expected memoryview");
         return NULL;
     }
 
-    if (PyObject_GetBuffer(Py_bytesarray_obj, &Py_buffer_obj, PyBUF_CONTIG) != 0) {
-        return NULL;
-    }
```
mnightingale, line 88: I'm not sure about this part, by just using PyMemoryView_GET_BUFFER further down we are presuming the backing is a bytearray and contiguous.

I'd suggest PyObject_CheckBuffer and PyObject_GetBuffer - could use PyBUF_CONTIG_RO case we ever wanted memoryview(bytes(...))
Would need to add the PyBuffer_Release at the end.

Alternatively just add a PyBuffer_IsContiguous check after PyMemoryView_GET_BUFFER.

tests/test_decoder.py
```diff
@@ -63,9 +63,9 @@ def test_end_after_filename():
         sabctools_yenc_wrapper(data_plain)
 
 
-def test_bad_size():
+def test_empty():
     with pytest.raises(ValueError) as excinfo:
-        sabctools.yenc_decode(bytearray())
+        sabctools.yenc_decode(memoryview(b""))
```
mnightingale, line 68: This should not matter, just highlighting that the test has changed and isn't what we'd usually call it with.
It is now testing a memoryview of bytes instead of bytearray.

I don't think there is anything wrong with allowing either.

src/yenc.cc
Safihre: I opted for PyBuffer_IsContiguous, a simple check.

tests/test_decoder.py
Safihre: True!

Safihre:
I also added the bytearray_malloc function, since it's so simple we can use it anyway!

mnightingale (approved):
Sorry I did have a look at the changes a couple of days ago, looks good to me.

Safihre:
Thanks :)

PR #114, opened 2024-05-08
Update yEnc module

animetosho:
This brings the yEnc code up to date with rapidyenc v1.1.1, and removes almost all sabctools' specific modifications (should allow easier updates in future).

## Key changes

* add AVX10 and experimental RISC-V Vector/crypto support
* add faster baseline CRC32 implementation (though crcutil still seems to be faster on older x86 processors)
* add platform optimised CRC manipulation routines, replacing such functionality from crcutil
  * this allows rapidyenc to optionally be built without crcutil
* avoid a performance pitfall on Zen4 processors
* officially support decoding where destination/source buffer are the same
  * fixes a rare bug with this setup, which could cause the end to be mis-detected
* de-dupe some code/lookup tables for a slight reduction in binary size

## Possible future changes

* I don't think [the `UNSUPPORTED_PLATFORM_ARM` check](https://github.com/animetosho/sabctools/blob/088ee38fc6d0894551b12d1b04477ff226bd1581/setup.py#L127-L133) is relevant any more, as the code itself checks for the header, but [it's left in](https://github.com/animetosho/sabctools/blob/088ee38fc6d0894551b12d1b04477ff226bd1581/src/yencode/common.h#L37-L39) as it might need some testing (this is the only remaining sabctools specific modification to rapidyenc)
* this PR redirects sabctools to use the new CRC functions, although I've kept crcutil present. If desired, crcutil can be disabled for a smaller build
* rapidyenc can now be built without the encoder, though sabctools does use it (only for testing I believe)
* there are [CRC32 tests where the CRC32 exceeds 32 bits](https://github.com/animetosho/sabctools/blob/088ee38fc6d0894551b12d1b04477ff226bd1581/tests/test_crc32.py#L13-L14). I'm not sure of the purpose of these tests (they don't appear to be related to #96), but I've left them in, except for a problematic one


sanderjo:
> experimental RISC-V Vector/crypto support

Wow! That is based on RV1.0? 

If so, time for me to buy a newer RISC-V SBC; my current one support RV0.7.x

sanderjo:
> I don't have any hardware

Well ... because you made the RV1.0 code ... I ordered RV1.0 hardware: "Banana Pi BPI-F3 SpacemiT K1 8 Core RISC-V chip".  To arrive at end of May.

Safihre:
The code changes are beyond my understanding, but do I understand correctly that crcutil is never used anymore? Even not as a fallback? Only if specifically enabled by compiler flag? Then indeed we can just remove it fully.

The crc32 exceeding 32bit, wasn't that for either the 32bit arm platforms or that exotic 32bit on 64bit platform that Linux supports?

The unsupported arm flag was for an old Synology chip, but I think we dropped that one already because unrar doesn't support old arm anymore. Will have to check that.

animetosho:
> but do I understand correctly that crcutil is never used anymore? Even not as a fallback? Only if specifically enabled by compiler flag?

It's still enabled by default, but only used on x86 CPUs which don't support PCLMULQDQ (which are quite old at this stage). There, the crcutil code is slightly faster than the fallback CRC code I have (e.g. 1150.71 vs 1325.75 MB/s on a Xeon L3426).

> The crc32 exceeding 32bit, wasn't that for either the 32bit arm platforms or that exotic 32bit on 64bit platform that Linux supports?

That doesn't quite sound right, as they should have no issue with sticking to 32-bit.  
My guess is that the crcutil interface took in 64-bit integers (since it was a generic CRC implementation, instead of being specific to CRC32), so it just got parsed as such. And because of that, the limits got tested that way.

mnightingale:
> there are [CRC32 tests where the CRC32 exceeds 32 bits](https://github.com/animetosho/sabctools/blob/088ee38fc6d0894551b12d1b04477ff226bd1581/tests/test_crc32.py#L13-L14). I'm not sure of the purpose of these tests (they don't appear to be related to [Incorrect CRC32 failure #96](https://github.com/sabnzbd/sabctools/issues/96)), but I've left them in, except for a problematic one

There is a bit of discussion around that in https://github.com/sabnzbd/sabctools/pull/80#issuecomment-1396275556 looks like an extreme case since Python would let you pass huge values and I think crcutil wouldn't handle them as expected. I believe we resolved to just disregard >32 bits.

---

Test failures are because macos-latest runner is an M1 (macos-14-arm64) then I think it tries to run x64 Python.
It should just need updating to use a x64 runner, and could have seperate arm64 added.

Safihre:
> It's still enabled by default, but only used on x86 CPUs which don't support PCLMULQDQ (which are quite old at this stage). There, the crcutil code is slightly faster than the fallback CRC code I have (e.g. 1150.71 vs 1325.75 MB/s on a Xeon L3426).

Seems like such a small benefit that we could just drop crcutil?
Or you still use it in rapidyenc also? Then we can keep it, just to keep things as similar as possible for easy future updates.

animetosho:
> There is a bit of discussion around that in #80 (comment) looks like an extreme case since Python would let you pass huge values and I think crcutil wouldn't handle them as expected. I believe we resolved to just disregard >32 bits.

I think that was for the length as opposed to the CRC32 itself?  
I did add tests to check the extremities of 64-bit lengths, since the new code does handle the full range correctly.

I suppose it doesn't matter much, but perhaps the tests involving >32-bit CRC32s can be removed, as they're not really a valid scenario.

> Test failures are because macos-latest runner is an M1 (macos-14-arm64)

Ooh, that's interesting. I've never been able to test anything on ARM Mac, so that could be useful if Github supports it now.

> Or you still use it in rapidyenc also? Then we can keep it, just to keep things as similar as possible for easy future updates.

I've left it enabled by default because it is technically faster and I don't really mind the size cost, but I support a build flag to disable it.  
If the size cost isn't a concern for you, it probably makes sense to keep it for now.

Safihre:
I fixed the macOS runs on the main branch by simply not forcing the architecture, only for the 32bit Windows build.

Thanks a lot for the update :)

PR #122, opened 2024-06-25
Fix building and linking OpenSSL on Windows ARM

### LINKED ISSUE #121: Build for Windows ARM64 fails
Building fails on Windows ARM64, I've narrowed it down to this.

https://github.com/sabnzbd/sabctools/blob/61121154f5bd3d9b0bf1dd5fa867c444061ab79c/setup.py#L346-L349

If I force the libraries line to run it builds and I can launch SAB on Windows 11 ARM.

But looking at it does it even need a platform check? It's already checking for `compiler_type == "msvc"` so it must be Windows AFAIK so can we just remove the IS_X86 part?

mnightingale:
Fixes #121

Public Arm runners are [coming soon](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/) so you will be able to replace the slow Linux QEMU build and add Windows Arm to the matrix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -314,7 +314,7 @@ class SABCToolsBuild(build_ext):
                     "src/unlocked_ssl.cc",
                 ],
                 "gcc_flags": ["-Wno-unused-parameter", "-Wno-missing-field-initializers"],
-                "msvc_x86_libraries": ["ws2_32"],
+                "msvc_libraries": ["ws2_32"],
             },
             {
                 "sources": [
@@ -345,8 +345,8 @@ class SABCToolsBuild(build_ext):
             if self.compiler.compiler_type == "msvc":
                 if IS_X86 and "msvc_x86_flags" in source_files:
                     args["extra_postargs"] += source_files["msvc_x86_flags"]
-                if IS_X86 and "msvc_x86_libraries" in source_files:
-                    ext.libraries += source_files["msvc_x86_libraries"]
+                if "msvc_libraries" in source_files:
+                    ext.libraries += source_files["msvc_libraries"]
             else:
                 if "gcc_flags" in source_files:
                     args["extra_postargs"] += source_files["gcc_flags"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -345,8 +345,8 @@ def build_extension(self, ext: Extension):
             if self.compiler.compiler_type == "msvc":
                 if IS_X86 and "msvc_x86_flags" in source_files:
                     args["extra_postargs"] += source_files["msvc_x86_flags"]
-                if IS_X86 and "msvc_x86_libraries" in source_files:
-                    ext.libraries += source_files["msvc_x86_libraries"]
+                if "msvc_libraries" in source_files:
```
Safihre, line 348: The lines above do need the x86 in them? Only these don't?

mnightingale: I could foresee a possible need to complement `msvc_x86_flags` with `msvc_arm_flags` in the future, and it keeps it simlar to how there are `gcc_flags`, `gcc_x86_flags`, `gcc_arm_flags` and others.

I don't think there would be any harm always including those flags and just rename it to `msvc_flags`, they seem to just have x86 relevant `/arch` flags so those codepaths are not used on arm anyway.
Maybe @animetosho can let us know when if they get a chance.

We we want the libraries linked regardless of architecture, so I thought it better to rename the key.

animetosho: I think using unsupported `/arch` flags causes MSVC to emit a warning and then ignore the issue. So you could set it on an ARM compile, but there's no reason to do so.

Safihre: 👍

Safihre:
@mnightingale Any chance you have time to see if you can get cross-compile working on Github Actions to get a arm64 wheel? 😇
cibuildwheel says they support it https://cibuildwheel.pypa.io/en/stable/faq/#windows-arm64

Safihre:
Oh wait, it's easy :) I'll do it!

PR #149, opened 2025-10-25
Add a streaming/incremental decoder

mnightingale:
Implements a streaming decoder to process NNTP traffic efficiently, especially important for pipelining because it can detect the end of responses.

Compatible with C++17 but uses C++20 features if available.

A `sabctools.Decoder()` instance holds the state of decoding, keep calling its decode method with a memoryview until it tells you it is done (`.\r\n` reached) but take care that it may also return data it couldn't process and you need to feed into a new Decoder.

The remaining/unprocessed data **must[^1]** be passed to a new `sabctools.Decoder` immediately. If that next decode fails then it needs moving to the start, and read more data from the network before calling decode again.

[^1]: the unprocessed data may contain a complete response; it needs processing because there may not be any further network reads for it.

### DIFF
--- a/src/yenc.cc
+++ b/src/yenc.cc
@@ -152,6 +152,8 @@ static inline bool extract_int(std::string_view line, const char* needle, T& des
  * @return Optional uint32_t with the parsed CRC, or nullopt if parsing fails
  */
 std::optional<uint32_t> parse_crc32(std::string_view crc32) {
+    if (crc32.empty()) return 0;
+
     uint64_t value = 0;
     auto [ptr, ec] = std::from_chars(crc32.data(), crc32.data() + crc32.size(), value, 16);
 
@@ -364,6 +366,7 @@ static inline void NNTPResponse_process_yenc_header(NNTPResponse* instance, std:
         instance->has_end = true;
         line.remove_prefix(5);
         std::string_view crc32;
+        bool crc32_found = false;
         // Multi-part files use pcrc32 (part CRC), single files use crc32
         constexpr std::string_view prefixes[] = { " pcrc32=", " crc32=" };
 
@@ -371,11 +374,12 @@ static inline void NNTPResponse_process_yenc_header(NNTPResponse* instance, std:
             auto pos = line.find(prefix, 5);
             if (pos != std::string::npos) {
                 crc32 = line.substr(pos + prefix.size());
+                crc32_found = true;
                 break;
             }
         }
 
-        if (crc32.size() >= 8) {
+        if (crc32_found) {
             instance->crc_expected = parse_crc32(crc32);
         }
 
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -282,3 +282,44 @@ def test_invalid_crc_chars():
     response = next(decoder, None)
     assert response
     assert response.crc_expected is None
+
+@pytest.mark.parametrize(
+    "hex,expected",
+    [
+        ["ffffffffa95d3e50", 0xa95d3e50],
+        ["fffffffa95d3e50", 0xa95d3e50],
+        ["ffffffa95d3e50", 0xa95d3e50],
+        ["fffffa95d3e50", 0xa95d3e50],
+        ["ffffa95d3e50", 0xa95d3e50],
+        ["fffa95d3e50", 0xa95d3e50],
+        ["ffa95d3e50", 0xa95d3e50],
+        ["fa95d3e50", 0xa95d3e50],
+        ["a95d3e50", 0xa95d3e50],
+        ["a95d3e5", 0xa95d3e5],
+        ["a95d3e", 0xa95d3e],
+        ["a95d3", 0xa95d3],
+        ["a95d", 0xa95d],
+        ["a95", 0xa95],
+        ["a9", 0xa9],
+        ["a", 0xa],
+        ["", 0],
+        ["12345678 ", 0x12345678],  # space at end
+    ],
+)
+def test_parsing_crc(hex: str, expected: int):
+    parts = [
+        b"222 0 <foo@bar>\r\n"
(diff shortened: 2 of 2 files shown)


Safihre:
Could you maybe also add example how we would use this? For dummies like me 😁

mnightingale:
> Could you maybe also add example how we would use this? For dummies like me 😁

Sure will do, getting pretty close now.

Currently I'm testing it by replacing the wrapper, only failure currently is test_ref_counts but I think that is expected because the Decoder will refer to the data_out.

I have to modify the input slightly because several yencfiles are not valid, e.g. test_end_after_filename.yenc doesn't have a \r\n or end with .\r\n

```py
def sabctools_yenc_wrapper(data: bytearray) -> Tuple[bytearray, str, int, int, int, Optional[int]]:
    return sabctools_yenc_streaming_wrapper(data)


def sabctools_yenc_streaming_wrapper(data: bytearray) -> Tuple[bytearray, str, int, int, int, Optional[int]]:
    decoder = sabctools.Decoder()

    done, remaining = decoder.decode(memoryview(data if data[-3:] == b".\r\n" else data + b"\r\n.\r\n"))
    assert done == True
    assert remaining == 0

    return decoder.data, correct_unknown_encoding(decoder.file_name), decoder.file_size, decoder.part_begin, decoder.part_size, decoder.crc
```

I havent included this in a commit because I'm not sure how we will handle it, remove the yenc_decode method?

mnightingale:
Eventually the goal will be a command queue of requests, send 5 requests, read back the responses in the same order.

```py
self.requests: Deque[sabnzbd.nzbstuff.Article] = deque()
self.responses: Deque[Tuple[sabnzbd.nzbstuff.Article, sabctools.Decoder]] = deque()
```

I haven’t really gotten into this part yet, I think we could do without responses and call straight into decode and eventually sabnzbd.NzbQueue.register_article - will need to handle that the response is already decoded.

The idea is that requests are what has been sent (pipeline), store self.article as what we are current reading so that's much unchanged, popleft() to get the next response to read once done is reached.

For stat recording I'll also need to add how much data was read to produce the decoded output.

thezoggy:
so it sounds like the goal is like
gWr51

but about returning the order..  wouldnt that slow things down trying to wait for all requests to be fullfilled to maintain order?

Safihre:
This turned out a lot more complicated than I thought.
I understand all parts of the code, just thought it would be simpler 😅

mnightingale:
I was afraid of scaring you off, I thought it best to keep it simple and start with just using it in a streaming way.

A queue of commands appears to require a lot of changes to the downloader/newswrapper logic, maybe a rethink of the overall design will be needed I think it will be well worth it eventually.

Head of line (HOL) blocking also needs to be a concern; I don’t know how we guard against it, maybe batching requests so they fit in one packet would help, request 5, wait for response of all 5, request another 5… 

It might be worthwhile implementing a pipelining POC like I did in Go, a single connection, then I think probably a command queue using deque, and threading.Semaphore to limit pending request.

I’ve also been looking at handling uu, it’s very simple so will probably add it. There are some areas I’ll refactor and other changes more focussed on efficiency, but I think it has everything important implemented.

Safihre:
I used Warp to add comments so I can better follow at my old age 😂

mnightingale:
I think I’ve overengineered this enough now, the latest change replaces the bool in the return tuple with a status which is also available as a property replacing success, and adds a property for the encoding format.

```py
status, remaining = decoder.decode(memoryview(data))
if status:
    # status >= 1 (SUCCESS, NO_DATA, INVALID_SIZE, etc.) or look at `decoder.status_code`
```

Checking if a complete response is available is still simple because they're all truthy, what to do with the response is still up to the calling application but it tries to categorise most status codes. Either use the status values, more complex logic with status_code, or manual checking decoder.crc etc.

I've left it as a draft as I probably wouldn’t merge it yet, but I don't plan on drastically changing the API anymore, unless something is found lacking when trying to integrate it into SAB.

Safihre:
So where do we go from here? :)

src/yenc.cc
```diff
@@ -23,225 +23,1053 @@
...
+ * 
+ * @param instance The Decoder instance to update
+ * @param line The line to examine for format detection
+ */
+static inline void decoder_detect_format(Decoder* instance, std::string_view line) {
+    // YEnc detection
+	if (starts_with(line, "=ybegin "))
+	{
+	    Py_XDECREF(instance->format);
+		instance->format = ENCODING_FORMAT_YENC;
```
Safihre: Why do we need the reference counting logic? It wil always already have count>0 because of the reference in the `yenc_init`?

mnightingale:
I think I can/will just make it return (bytearray, memoryview).

thezoggy:
> Loads of changes to get it to work as an iterator, but as expected it's much nicer to use from Python now.
> 
> ```python
> # Receive data into the decoder pre-allocated buffer
> if not nbytes and self.nntp.nw.server.ssl and not self.nntp.nw.blocking and sabctools.openssl_linked:
>     # Use patched version when downloading
>     bytes_recv = sabctools.unlocked_ssl_recv_into(self.nntp.sock, self.decoder)
> else:
>     bytes_recv = self.nntp.sock.recv_into(self.decoder, nbytes=nbytes)
> 
> self.decoder.process(bytes_recv)
> for response in self.decoder:
>     with self.lock:
>         article = self._response_queue.popleft()
>     self.command_complete(response, article)
> ```
> 
> I'll also be more efficient because for each read it doesn't have to go back and forth Python-C. The biggest issue I'm seeing is the assembler can't keep up much beyond 10-20Gbps, I have to limit it while testing. I have several ideas to improve that so expect a PR sometime.

just curious how you are testing those speeds, local nserve using ramdrive?

PR #150, opened 2025-10-26
Replace pickles with yenc files

mnightingale:
I find it useful to be able to easily open them in a text editor to figure out what went wrong.

I unpickled them with:

```py
import glob
import os
from pathlib import Path

from tests.testsupport import read_pickle, read_plain_yenc_file

for fname in glob.glob("tests/yencfiles/*.pickle"):
    data_plain = read_pickle(fname)

    dest = Path(fname).with_suffix(".yenc")
    with open(dest, 'xb') as fw:
        fw.write(data_plain)

    assert data_plain == read_plain_yenc_file(dest.name)

    os.remove(fname)
```

### DIFF
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -105,14 +105,14 @@ def test_ref_counts():
     assert sys.getrefcount(data_plain) == expected_refcount
 
 
-def test_crc_yencs():
+def test_crc_yenc_files():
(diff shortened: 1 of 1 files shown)


Safihre:
Yeah but the point of the pickles was more about that they contain the raw chunks. So split how they would come of the socket.

Safihre:
Indeed doesn't do that anymore. 
However, they do contain a bit more real world random splitting of data.

PR #153, opened 2025-11-06
Increase minimum to C++17 but use C++20 if available

mnightingale:
I wanted to use the autoconf_check method on Windows, but I discovered it doesn't work because `tempfile.NamedTemporaryFile` has issues, on Windows the permissions it sets do not allow reopening of files.
So calling the compiler fails because it can't read the file.

You can do the following, but then you have to mess around deleting the file after in multiple exit paths.

```py
f = tempfile.NamedTemporaryFile("w", suffix=".cc", delete=False)
#...
f.close()
#...
compiler.compile(...)
```

Another issue I found is that it places the output in the working directory (sabctools\Users\Administrator\AppData\Local\Temp), it deletes the output file but not the directories.

To resolve the issues I instead create a temporary directory to use for the .cc and .obj files, and contextlib.ExitStack to cleanup using shutil.rmtree.

---

My actual intention was bumping the minimum to C++17 but using C++20 if available, so I've added checks for supported C++ versions and fixed a couple of minor errors C++20 on (goto statements which jump initialisation).

---

I noticed 3.8 has some warnings, may need pinning to an older setuptools version for 3.8 or drop support for it like SAB has?

Link requires a different Python (3.8.18 not in: '>=3.9'): https://files.pythonhosted.org/packages/21/df/7c6bb83dcb45b35dc35b310d752f254211cde0bcd2a35290ea6e2862b2a9/setuptools-75.4.0-py3-none-any.whl (from https://pypi.org/simple/setuptools/) (requires-python:>=3.9)

### DIFF
--- a/.github/workflows/ci_and_build_release.yml
+++ b/.github/workflows/ci_and_build_release.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14' ]
+        python-version: [ '3.9', '3.10', '3.11', '3.12', '3.13', '3.14' ]
 
     steps:
       - uses: actions/checkout@v5


Safihre:
We can drop 3.8.

PR #160, opened 2025-12-23
Allocate files optionally with preallocation and sparse

mnightingale:
This makes a couple of changes which allow the caller to choose the behaviour they need:

## Windows

Argument for `sparse` default True, if article cache is disabled writes would be pretty random resulting in heavily fragmented files.
i.e. `Article{1,5,2,4,3}` which is undesirable, so in that case it is best not to make the file sparse and only update the logical size of the file - no data is written but it reserves the size on disk and Windows may be able to make smarter decisions for random writes. So writing 1 then 5 may requiring writing zeros, but then 2,4,3 can fill in the gap and hopefully prevent creating fragmented files.

Obscure filesystems are less likely on Windows so I didn't implement manual preallocation.

## POSIX

Argument `preallocate` default True, uses `posix_fallocate` if available (Linux) to reserve disk space, most file systems won't actually write zeros to disk just update metadata.

If not available falls back to `ftruncate` to create a sparse file, usually no immediate writes except FAT/exFAT.

If both fail, falls back to manually preallocate the file by writing `\0` which could be slow, but it would be very rare obscure filesystems, things like some custom FUSE FS, maybe older NFS or SMB servers.

## How it will be used

It will be used as `allocate_file(fd, length, sparse=cache_enabled, preallocate=not cache_enabled)` (maybe also check for FAT).

On Windows we mostly benefit from sparse files if we have the article cache since we will usually write sequentially and can leave gaps for missing articles, so preallocation has little effect.

On POSIX we only need to preallocate if there is no article cache (rare) and it attempts various techniques, I could probably have gone without the manual fallback it would just result in fragmented files but to get there you need a very odd setup.

The defaults most user will be passing are `sparce=True, preallocate=False` with no zeroing of files, sequential writes, and efficient gaps. They'll be able to disable direct write to operate in the current append mode.


Safihre:
Why even support not using it? The arguments don't seem super strong?

mnightingale:
> Why even support not using it?

Sorry which do you mean?

With Windows, FAT/exFAT doesn't support sparse files and setting the file size will preallocate the file (slowly write lots of data), NTFS is not a problem.

The main reason for making sparse configurable is because if the article cache is not enabled but sparse files were allowed, writes would produce very fragmented files. Because in direct write mode I'm cutting out the temporary write/read/write cycle and writing direct to the file, so sparse files on Windows would mostly append the data and store metadata for the order of parts. So an option to turn sparse off on Windows I think is quite necessary.

Preallocation on the other hand on POSIX I don't think is so important, I could remove the `posix_fallocate` or at least manual allocation. Just keep `ftruncate` and those rare filesystems that don't support it might produce more fragmented files.
On non-CoW filesystems `posix_fallocate` might reduce fragmentation, ext4 allocates (slow), XFS instant, but I don't think the risk/tradeoff is worth it when not using it just means more fragmented files and users should just enable the cache - even a small size would allow more sequential writes.

In hopefully most configurations, users have left the cache enabled so writes will usually be sequential resulting in not very fragmented files and these days many will be using SSDs for incomplete so fragmented files are not an issue at all.

mnightingale:
I think it will work as is if it'll only be used if the cache is enabled.

PR #161, opened 2025-12-23
Validate yenc size and offsets

mnightingale:
Restored the previous validation except `part_end > part_begin` because the `part_size > 0` check already covers that (part_size can't be > 0 and part_end <= part_begin).

Allows us to check for valid offsets with `part_begin and part_end` since on failure they'd both be 0, likewise non-multipart will always be 0,0 so always append mode for writing.

https://github.com/sabnzbd/sabctools/blob/a38c4c1c85aa79ef5e8edef0297b47c96ee4f0ab/src/yenc.cc#L144-L160


Safihre:
I think we should at least add 1 test for this.

PR #162, opened 2026-01-05
Implement __bool__ and __len__ on Decoder

mnightingale:
For a change fixing something it'll be nicer to be able to do:

```py
if self.decoder:
    # process responses
    # maybe stop listening for socket events
```

Rather than:

```py
has_response = False
for response in self.decoder:
    has_response = True
    # process responses

if has_response:
    # maybe stop listening for socket events
```

I'll break one place we do `if not self.decoder` but I'll fix that in the PR.

Behaviour is the same as you wuld get from a list, set, or other collection container.

### DIFF
--- a/src/yenc.cc
+++ b/src/yenc.cc
@@ -1210,6 +1210,17 @@ static void Decoder_dealloc(Decoder *self)
     Py_TYPE(self)->tp_free((PyObject*)self);
 }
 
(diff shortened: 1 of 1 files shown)


mnightingale:
Yes, I'd say so since it is a behaviour change of `if decoder:`

PR #178, opened 2026-05-10
Fix licensing

scaronni:
Avoid licensing error when building:

```
ValueError: invalid pyproject.toml config: `project.license`.
configuration error: `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
```


Safihre:
You are running an old setuptools, see #171

Safihre:
I have now made it an explicit requirement to have modern setuptools.

PR #185, opened 2026-07-02
Fix decoder memory leaks

mnightingale:
Some `Py_XDECREF` are probably not strictly necessary (file_name and message), but I think better to be there in case they can happen with malformed articles.

The problem ones are NNTPResponse message and lines.
Decoder  response.
`NNTPResponse_dealloc` changed order to match the members.
`Decoder_init` allocates the deque first, so it's not leaked if out of memory.

https://github.com/sabnzbd/sabnzbd/issues/3473

### DIFF
--- a/src/yenc.cc
+++ b/src/yenc.cc
@@ -357,6 +357,7 @@ static inline void NNTPResponse_process_yenc_header(NNTPResponse* instance, std:
 	        line.remove_prefix(pos + 6);
             // Strip trailing whitespace/null from filename
             if ((pos = line.find_last_not_of('\0')) != std::string::npos) {
+                Py_XDECREF(instance->file_name);
                 instance->file_name = decode_utf8_with_fallback(line.substr(0, pos + 1));
             }
 	    }
@@ -420,7 +421,6 @@ static void NNTPResponse_append_line(NNTPResponse* instance, std::string_view li
 
     if (instance->lines == nullptr) {
         instance->lines = PyList_New(0);
-        Py_INCREF(instance->lines);
     }
 
     PyList_Append(instance->lines, py_str);
@@ -436,9 +436,10 @@ static void NNTPResponse_dealloc(NNTPResponse* self)
 {
     Py_XDECREF(self->decoder);
     Py_XDECREF(self->data);
-    Py_XDECREF(self->file_name);
     Py_XDECREF(self->lines);
     Py_XDECREF(self->format);
+    Py_XDECREF(self->file_name);
+    Py_XDECREF(self->message);
     Py_TYPE(self)->tp_free((PyObject*)self);
 }
 
@@ -739,6 +740,7 @@ static bool NNTPResponse_decode_uu(NNTPResponse* instance, std::string_view line
             trim_while(line, [](const unsigned char c){ return std::isspace(c); }); // skip spaces after permissions
 
             // The rest of the line is the filename
+            Py_XDECREF(instance->file_name);
             instance->file_name = decode_utf8_with_fallback(line);
 
             instance->body = true;
@@ -887,6 +889,7 @@ static Py_ssize_t NNTPResponse_decode_buffer(NNTPResponse *instance, const char*
         if (instance->format == nullptr) {
             if (!instance->status_code && line.length() >= 3) {
                 // Store the full command response line
+                Py_XDECREF(instance->message);
                 instance->message = decode_utf8_with_fallback(line);
                 // First line should be NNTP status code (220, 222, 223, etc.)
                 if (!extract_int(line, "", instance->status_code)
@@ -1180,16 +1183,17 @@ static int Decoder_init(Decoder *self, PyObject *args, PyObject *kwds)
     if (size > YENC_MAX_PART_SIZE)
         size = YENC_MAX_PART_SIZE;
 
+    new (&self->deque) std::deque<NNTPResponse*>();
     self->response = nullptr;
     self->data = static_cast<char *>(malloc(size));
     self->size = size;
     self->consumed = 0;
     self->position = 0;
     if (!self->data) {
+        self->deque.~deque();
         PyErr_NoMemory();
         return -1;
     }
-    new (&self->deque) std::deque<NNTPResponse*>();
 
     return 0;
 }
@@ -1207,6 +1211,7 @@ static void Decoder_dealloc(Decoder *self)
     // DECREF all remaining items
     for (NNTPResponse* item : self->deque)
         Py_XDECREF(item);
+    Py_XDECREF(self->response);
     self->deque.~deque();
     free(self->data);
     Py_TYPE(self)->tp_free((PyObject*)self);


Safihre:
> [!NOTE]
> This analysis was written by **Claude Fable** (Anthropic's `claude-fable-5` model), asked to audit these files for further leaks after this PR was merged. It is based on reading the code, not on running a leak checker — **please verify each point before acting on it.**

First: the fixes in this PR look complete for what they target. All six `PyObject*` members of `NNTPResponse` (`decoder`, `data`, `lines`, `format`, `file_name`, `message`) are now released in `NNTPResponse_dealloc`, and `Decoder_dealloc` now covers `self->response`. The findings below are *additional* issues in `yenc.cc`, `sparse.cc` and `unlocked_ssl.cc`, ordered by impact.

## yenc.cc

### 1. Uncollectable `Decoder` ⇄ `NNTPResponse` reference cycle (biggest remaining leak)

Every `NNTPResponse` takes a strong reference to its parent `Decoder` ([yenc.cc#L1011-L1012](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1011-L1012)), while the `Decoder` holds strong references back via `self->response` and the deque. Neither type sets `Py_TPFLAGS_HAVE_GC` or implements `tp_traverse`/`tp_clear` ([yenc.cc#L1101](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1101), [yenc.cc#L1407](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1407)), so Python's cyclic GC can never see or break the cycle.

Consequence: if a `Decoder` is dropped while an article is still in flight (`self->response != nullptr`, e.g. a connection aborted mid-article) or while responses are still queued in the deque, the decoder, its malloc'd input buffer, and every pending response (with its bytearray, up to 10 MB) leak **permanently**. Only a fully-drained decoder is reclaimable. In normal SABnzbd operation (connections being reset/recycled) this can leak megabytes per dropped connection.

Fix options: add GC support to both types, or drop/weaken the back-reference — it appears to exist only to serve `NNTPResponse_iternext` (see next point).

### 2. `NNTPResponse_iternext` leaks one reference per yielded item

At [yenc.cc#L931-L933](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L931-L933): `pop_front()` discards the deque's owned reference without a `Py_DECREF`, then `Py_INCREF(item)` adds another on top — net one reference nobody owns, so every response obtained via `next(response)` is immortal (and pins the decoder through its back-reference). Compare `Decoder_iternext` ([yenc.cc#L1164-L1173](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1164-L1173)), which correctly transfers ownership without the extra INCREF. The `Py_INCREF` should be deleted — or the whole function removed if it's vestigial (the type has no `tp_iter`, so it only fires on a direct `next()` call).

### 3. `Decoder_init` is not safe against being called twice

`tp_init` is re-invokable from Python (`d.__init__(8192)`). A second call placement-news over the live deque at [yenc.cc#L1186](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1186) (leaking its heap nodes and the owned response references), nulls `self->response` without a DECREF, and overwrites `self->data` without `free()`. Same bug class as the `Decoder_init` ordering issue fixed in this PR; needs a guard, or the one-time construction moved to `tp_new`.

### 4. Minor (OOM-only or module-init-only paths)

- [yenc.cc#L422-L426](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L422-L426) — if `PyList_New(0)` fails, `PyList_Append(nullptr, ...)` dereferences NULL (crash) and `py_str` leaks; a `PyList_Append` failure is also silently swallowed, leaving a pending exception.
- [yenc.cc#L1455-L1460](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1455-L1460) — `create_int_enum` never releases the `PyLong_FromLong` reference it passes to `add_member`, so each enum value ends up one reference high. Harmless in practice (0 and 1 are cached small ints), but the ownership contract is wrong.
- [yenc.cc#L1492-L1508](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L1492-L1508) — if `ENCODING_FORMAT_YENC` resolves but `UU` fails, the YENC reference leaks; the `PyModule_AddObject` partial-failure handling also DECREFs references that were already stolen.

Not a bug, in case anyone wonders: the double `PyBuffer_Release(&dst_buf)` in the resize-failure path ([yenc.cc#L622-L627](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/yenc.cc#L622-L627)) is a no-op the second time because the first release NULLs `view->obj`.

## sparse.cc

Clean — every temporary is released on both success and error paths. Only nit: `Py_file_truncate` ([sparse.cc#L39](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/sparse.cc#L39)) is never assigned anywhere — dead code.

## unlocked_ssl.cc

No leaks found, but two adjacent refcount/correctness issues:

- **Latent over-DECREF of a borrowed reference** (the opposite of a leak — worse): `sock` comes from `PyWeakref_GetObject`, which returns a *borrowed* reference, and is only INCREF'd at [unlocked_ssl.cc#L188](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/unlocked_ssl.cc#L188). The two early exits before that (`goto error` at L175, `goto done` at L179) land on `Py_XDECREF(sock)` — decrementing a reference that was never taken, which would prematurely free the socket. Currently unreachable because the only caller guarantees `0 < len == buffer->len`, but CPython's original `_ssl.c` does the INCREF *before* those length checks for exactly this reason.
- [unlocked_ssl.cc#L278](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/unlocked_ssl.cc#L278) — `blocking = PyObject_CallMethod(...)` has no NULL check; if the call raises, execution continues with a pending exception set.
- Hygiene: the `openssl_init` failure cleanup ([unlocked_ssl.cc#L144-L148](https://github.com/sabnzbd/sabctools/blob/a0ae236/src/unlocked_ssl.cc#L144-L148)) DECREFs the globals but doesn't NULL them, leaving dangling pointers. They are only ever truth-tested afterwards, so currently harmless. (The kept `dlopen` handle on the POSIX path is intentional, not a leak.)

---
*Line links point at master (a0ae236). Again: this is an AI-generated review — please double-check each finding before filing fixes.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mnightingale:
I pretty much agree with the assessment, however I don’t think it’s as serious as it makes out because it’s not aware of what the Python side does; all responses of a decoder are drained. Regardless, I don’t think we are even using`NNTPResponse_iternext` so it can be removed to resolve 1 and 2.

3 for sure is an issue but we never hit it, I will probably just prevent reinit because it would break internal state, i.e. would it shrink a buffer with partial data or clear the deque? So simplest just to return an exception if it’s attempted and perhaps also move init to tp_new.

4 as it says are minor and we already knew sabctools leaks globals when imported; something we might need to fix for free threading eventually anyway. So, I might just fix the first and last bullet point for now.

Sparse minor cleanup of dead code

Unlocked, I think I’ll PR the 3.15 changes and we can reassess after because the ` PyWeakref_GetObject` replacement doesn’t return a borrowed reference, so it changes things, but we have the complication of supporting both. The other issues are fairly minor.

PR #187, opened 2026-07-03
Fix leaks in yenc and sparse dead code

mnightingale:
Fixes issues in yenc.cc and sparse.cc listed on https://github.com/sabnzbd/sabctools/pull/185#issuecomment-4874054872

- Remove `NNTPResponse` references to parent `Decoder` and remove `NNTPResponse` as an iterator
- Prevent `Decoder` being initialized multiple times and leaking
- Minor OOM or module-init issues
  - Fix PyList_New(0) failure crash and leak
  - Fix `create_int_enum` ownership contract paths, `PyLong_FromLong` not decref'd
  - Fix module init paths leaking `ENCODING_FORMAT_*`
- Removed no-op call to `PyBuffer_Release`
- Sparse remove dead code

Maybe Claude could check?

I'll PR unlocked_ssl separately, I've not looked at it yet but I don't think it needs a release until after that.

### DIFF
--- a/src/sparse.cc
+++ b/src/sparse.cc
@@ -36,7 +36,6 @@ PyObject *sparse(PyObject *self, PyObject *args)
 
     PyObject *Py_file_fileno = NULL;
     PyObject *Py_file_handle = NULL;
-    PyObject *Py_file_truncate = NULL;
 
     // Accept either a file object (with fileno()) or an integer file descriptor directly
     if (!PyArg_ParseTuple(args, "OL:sparse", &Py_file, &length))
@@ -125,12 +124,10 @@ PyObject *sparse(PyObject *self, PyObject *args)
 
     Py_XDECREF(Py_file_fileno);
     Py_XDECREF(Py_file_handle);
-    Py_XDECREF(Py_file_truncate);
     Py_RETURN_NONE;
 
 error:
     Py_XDECREF(Py_file_fileno);
     Py_XDECREF(Py_file_handle);
-    Py_XDECREF(Py_file_truncate);
     return NULL;
 }
--- a/src/yenc.cc
+++ b/src/yenc.cc
@@ -1437,20 +1437,6 @@ struct EnumEntry {
     long value;
 };
 
-// Add a key, value to a dictionary
-static bool add_member(PyObject* dict, const char* name, PyObject* value) {
-    if (!dict || !name || !value) return false;
-
-    Py_INCREF(value);
-    if (PyDict_SetItemString(dict, name, value) < 0) {
-        Py_DECREF(value);
-        return false;
-    }
-    Py_DECREF(value);
-
-    return true;
-}
-
 // Create an int enum for a list of entries
 static PyObject* create_int_enum(const char* enum_name, const EnumEntry* entries, std::size_t count) {
     if (!enum_name || !entries) return nullptr;
@@ -1461,10 +1447,20 @@ static PyObject* create_int_enum(const char* enum_name, const EnumEntry* entries
     // Range over the entries
     for (std::size_t i = 0; i < count; ++i) {
         const auto& e = entries[i];
-        if (!add_member(members, e.name, PyLong_FromLong(e.value))) {
+
+        PyObject* value = PyLong_FromLong(e.value);
+        if (!value) {
             Py_DECREF(members);
             return nullptr;
         }
+
+        if (PyDict_SetItemString(members, e.name, value) < 0) {
+            Py_DECREF(value);
+            Py_DECREF(members);
+            return nullptr;
+        }
+
+        Py_DECREF(value);
     }
 
     PyObject* enum_module = PyImport_ImportModule("enum");
--- a/src/yenc.cc
+++ b/src/yenc.cc
@@ -414,6 +414,7 @@ static inline void NNTPResponse_process_yenc_header(NNTPResponse* instance, std:
  *
  * @param instance Decoder instance whose lines list will be appended to.
  * @param line     Line contents without the trailing CRLF.
+ * @return 0 on success, or -1 on error
  */
 static int NNTPResponse_append_line(NNTPResponse* instance, std::string_view line) {
     auto py_str = decode_utf8_with_fallback(line);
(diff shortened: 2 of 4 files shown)


Safihre:
> [!NOTE]
> Review by **Claude Fable** (Anthropic's `claude-fable-5` model), as a follow-up to the [audit on #185](https://github.com/sabnzbd/sabctools/pull/185#issuecomment-4874054872). Multi-angle automated review of this PR's diff — please verify before acting.

**Overview: the PR correctly fixes every issue it targets.** Specifically verified:

- The `Decoder` ⇄ `NNTPResponse` cycle is fully resolved by removing the back-reference — no remaining cycles exist (`NNTPResponse` now only holds bytearray/list/str/enum members), and the deque's ownership transfer in `Decoder_iternext` is balanced (one ref from `PyObject_New` → deque → caller, with `Decoder_dealloc` covering leftovers).
- The reinit guard plus moving deque construction to `tp_new` makes every lifecycle path safe: a `__new__`-only object is now safe in all slots (zero-length memoryview, `process()` raises `ValueError`, dealloc valid), and a failed first `__init__` leaves a cleanly retriable state.
- The `yenc_init` restructure handles `PyModule_AddObject`'s steal-on-success-only contract correctly, no `goto` crosses a variable initialization (well-formed C++), and the globals are nulled on failure. `create_int_enum`'s refcounting is now correct.
- The removed `PyBuffer_Release` really was a redundant no-op (the earlier release NULLs `view->obj`); the sparse.cc deletion is pure dead code.
- The two intentional API changes (reinit raises `RuntimeError` — tested; `next(response)` now raises `TypeError`) break nothing in-repo: the `.pyi` never declared `NNTPResponse.__next__`.

Nothing blocking. Six polish-level findings, most fixable in one line, most severe first:

### 1. `NNTPResponse_init` keeps a now-unused `parent` parameter and a stale docstring

[yenc.cc#L996-L1006](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L996-L1006) — the doc comment still says the function "Binds the NNTPResponse to its owning Decoder" and that `parent` "will be referenced for iteration"; the call site at [yenc.cc#L1268](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L1268) still passes the decoder. A future maintainer trusting the doc could re-add `Py_INCREF(parent)` — and since `NNTPResponse_dealloc` no longer releases any decoder field, that single change would silently reintroduce a permanent `Decoder` leak (worse than the original cycle: no back-ref field means nothing would ever DECREF it). Suggest dropping the parameter, the argument, and the two doc lines.

### 2. `// empty lines are ignored` mislabels the decode-failure case

[yenc.cc#L421-L422](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L421-L422) — `decode_utf8_with_fallback` returns `nullptr` both for empty lines **and** when decoding fails (OOM-only, since Latin-1+replace can't fail on content), clearing the error internally. That OOM silently drops a non-empty line from `response.lines` and reports success, while the identical OOM one statement later (`PyList_New`/`PyList_Append`) now correctly returns -1 and raises. The silent drop itself matches pre-existing `file_name`/`message` behavior, so at minimum fix the comment (e.g. `// empty or undecodable lines are skipped`); ideally distinguish the two cases.

### 3. `PyErr_SetNone(PyExc_StopIteration)` replaces the bare-NULL exhaustion signal on a hot path

[yenc.cc#L1159-L1162](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L1159-L1162) — the `tp_iternext` contract explicitly allows returning NULL with *no* exception set on exhaustion ("a StopIteration exception may or may not be set"), and that is the fast path: `next(decoder, None)` and `FOR_ITER` skip the exception match+clear entirely. The consumer pattern is `next(decoder, None)` after every `process()` call, so the empty branch fires once per network read, and on CPython ≥3.12 `PyErr_SetNone` materializes an exception instance that the caller immediately matches, clears, and frees. Reverting to the old bare `return NULL;` is observably identical and free. (Tens of nanoseconds — a nit, but a one-line one.)

### 4. Awareness only: a failed `append_line` mid-pass can duplicate recorded lines on retry

[yenc.cc#L917-L919](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L917-L919) — the new `-1` propagates to `Decoder_process` before `self->consumed` advances, so lines already appended to `response.lines` in this pass would be re-parsed and re-appended if a caller caught the `MemoryError` and kept feeding the same decoder. OOM-only trigger, and the same non-transactional hazard already existed on master via the `decode_uu`/`decode_yenc` failure paths — flagging for awareness, not action.

### 5. The two hand-rolled type registrations are exactly `PyModule_AddType()`

[yenc.cc#L1499-L1512](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L1499-L1512) — `PyModule_AddType(m, &DecoderType)` / `PyModule_AddType(m, &NNTPResponseType)` (CPython ≥3.9; project floor is 3.10) does the INCREF/add/DECREF-on-failure dance internally and registers under the last component of `tp_name`, replacing ~10 lines. Incidentally: `DecoderType`'s `tp_name` is `"sabctols.Decoder"` — pre-existing typo ("sabctols"), harmless since only the last component is used, but worth fixing while in the file.

### 6. The error label's XDECREF+null pairs are `Py_CLEAR`

[yenc.cc#L1523-L1526](https://github.com/sabnzbd/sabctools/blob/7499bba/src/yenc.cc#L1523-L1526) — `Py_CLEAR(ENCODING_FORMAT_YENC); Py_CLEAR(ENCODING_FORMAT_UU);` is the canonical idiom: two statements instead of four, and it clears the pointer before decrefing.

---
*Also considered and rejected, so nobody re-litigates: a "second `yenc_init` failure clobbers globals a prior init owns" scenario — refuted, since this is a single-phase module (`m_size = -1`) whose init never re-runs while a prior instance is live, and master behaves no better in that hypothetical.*

*Line links point at 7499bba. AI-generated review — please double-check before acting.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Safihre:
Claude approves 😁

mnightingale:
I wrote and must have forgotten to submit a comment yesterday, the remaining minor issues were addressed. I think a few are Python 3.10 things we are now able to use.
I have a bunch of other changes to make which mostly cleanup how we store state, create type definitions, support unloading and subinterpreters, but I'll send them in another PR.

PR #188, opened 2026-07-05
Python 3.15 support

mnightingale:
Fixes some issues identified in https://github.com/sabnzbd/sabctools/pull/185#issuecomment-4874054872

- Early exits; no space in buffer or too large
- Type check that we are given an SSLSocket
- getblocking check accounts for call failure
- Adds 3.15 support using PyWeakref_GetRef on 3.13+
- openssl_init single path, remove dlclose, and cleanup to prevent dangling pointers

### DIFF
--- a/src/unlocked_ssl.cc
+++ b/src/unlocked_ssl.cc
@@ -324,7 +324,7 @@ PyObject* unlocked_ssl_recv_into(PyObject* self, PyObject* args) {
     }
 
     Py_ssl_socket = PyObject_GetAttrString(ssl_socket, "_sslobj");
-    if (!Py_ssl_socket) {
+    if (!Py_ssl_socket || Py_IsNone(Py_ssl_socket)) {
         PyErr_SetString(PyExc_ValueError, "Could not find _sslobj attribute");
         goto error;
     }
--- a/src/unlocked_ssl.cc
+++ b/src/unlocked_ssl.cc
@@ -208,16 +208,13 @@ void openssl_init() {
 #endif
 
 cleanup:
-    Py_XDECREF(ssl_module);
-    Py_XDECREF(_ssl_module);
-    Py_XDECREF(_ssl_module_path);
+    Py_CLEAR(ssl_module);
+    Py_CLEAR(_ssl_module);
+    Py_CLEAR(_ssl_module_path);
     if (!openssl_linked()) {
-        Py_XDECREF(SSLWantReadError);
-        Py_XDECREF(SSLWantWriteError);
-        Py_XDECREF(SSLSocketType);
-        SSLWantReadError = NULL;
-        SSLWantWriteError = NULL;
-        SSLSocketType = NULL;
+        Py_CLEAR(SSLWantReadError);
+        Py_CLEAR(SSLWantWriteError);
+        Py_CLEAR(SSLSocketType);
         openssl_handle = NULL;
         SSL_read_ex = NULL;
         SSL_get_error = NULL;
--- a/src/unlocked_ssl.cc
+++ b/src/unlocked_ssl.cc
@@ -330,6 +330,7 @@ PyObject* unlocked_ssl_recv_into(PyObject* self, PyObject* args) {
     Py_buffer Py_buffer;
     PyObject *retval = NULL;
     PyObject *blocking = NULL;
+    int is_blocking;
 
     if(!openssl_linked()) {
         PyErr_SetString(PyExc_OSError, "Failed to link with OpenSSL");
@@ -348,7 +349,12 @@ PyObject* unlocked_ssl_recv_into(PyObject* self, PyObject* args) {
     }
 
     blocking = PyObject_CallMethod(ssl_socket, "getblocking", NULL);
-    if (blocking == Py_True) {
+    if (!blocking)
+        goto error; // call failed
+    is_blocking = PyObject_IsTrue(blocking);
+    if (is_blocking < 0)
+        goto error;
+    if (is_blocking) {
         PyErr_SetString(PyExc_ValueError, "Only non-blocking sockets are supported");
         goto error;
     }
--- a/src/unlocked_ssl.cc
+++ b/src/unlocked_ssl.cc
@@ -153,8 +153,6 @@ static int get_socket(PySSLSocket *obj, PySocketSockObject **out_sock)
 
 /* Linking to OpenSSL function used by Python */
 void openssl_init() {
-    // TODO: consider adding an extra version check to avoid possible future changes to SSL_read_ex
-
     PyObject *ssl_module = NULL;
     PyObject *_ssl_module = NULL;
     PyObject *_ssl_module_path = NULL;
@@ -197,7 +195,9 @@ void openssl_init() {
     if (PyObject_HasAttrString(_ssl_module, "__file__")) {
         _ssl_module_path = PyObject_GetAttrString(_ssl_module, "__file__");
         if(!_ssl_module_path) goto error;
-        openssl_handle = dlopen(PyUnicode_AsUTF8(_ssl_module_path), RTLD_LAZY | RTLD_NOLOAD);
+        const char *path = PyUnicode_AsUTF8(_ssl_module_path);
+        if(!path) goto error;
+        openssl_handle = dlopen(path, RTLD_LAZY | RTLD_NOLOAD);
     }
 
     if(!openssl_handle) goto error;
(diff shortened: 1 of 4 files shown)


Safihre:
Claude had 1 more small thing on the previous PR:

One tiny behavior nuance worth a glance, not a defect: the empty-line special case was removed from decode_utf8_with_fallback itself (it now decodes empty input to "" rather than returning nullptr). Every call site is unaffected except one edge: a malformed UU begin line with no filename now produces file_name == "" where it previously produced None. Both are falsy, so if response.file_name: code behaves identically; only an explicit is None check would notice. Probably fine — arguably more consistent — but it's the only observable change the follow-up commits introduced beyond the fixes.

src/unlocked_ssl.cc
```diff
@@ -269,14 +344,24 @@ PyObject* unlocked_ssl_recv_into(PyObject* self, PyObject* args) {
         return NULL;
     }
 
+    if (!PyObject_TypeCheck(ssl_socket, SSLSocketType)) {
+        PyErr_Format(PyExc_TypeError, "argument 1 must be %s", SSLSocketType->tp_name);
+        return NULL;
+    }
```
Safihre: This check is dead code: `PyArg_ParseTuple` with `O!` already performs exactly `PyObject_TypeCheck` against `SSLSocketType` and raises the same TypeError (that's what `test_unlocked_ssl_recv_into_not_a_socket_fails` asserts on), so this can never fire.

And if it ever did become reachable (e.g. `O!` later relaxed to `O`), the `return NULL` would skip `PyBuffer_Release`, leaking the `w*` buffer — it would need `goto error`.

Suggest deleting the block; the `O!` format plus the now-correct `PyTypeObject*` typing already covers the type check.

Safihre (commented):
> [!NOTE]
> This review was written by **Claude** (Anthropic's `claude-fable-5` model), asked to review this PR. It is based on reading the code, cross-checking CPython sources, and building/testing the branch locally on Windows — but please verify each point before acting on it.

## Verification done

- **The 3.15 struct change is real and the `#if` guard is necessary.** In CPython main ([gh-148292](https://github.com/python/cpython/issues/148292)), `_PySSLError err` and `PyObject *exc` were replaced by `int got_eof_error`. Without the guard, writing `self->err` on 3.15 would clobber `got_eof_error` and write past the end of the real struct. The declared struct prefix (`Socket`, `ssl`, `ctx`, …) is unchanged across 3.10–main, and `sock_fd` is the first field of `PySocketSockObject` in all versions, so the `INVALID_SOCKET` check reads the right memory.
- **`get_socket()` mirrors CPython main's own `get_socket()`** almost line-for-line (minus the `BIO_set_nbio` resync, which sabctools can't do — acceptable since blocking sockets are rejected up front).
- **Builds clean** on MSVC / Python 3.14. 7/8 tests in `test_unlocked_ssl.py` pass; `test_unlocked_ssl_recv_into_bulk_response` fails identically on master on the same machine (a `sendall` flake on non-blocking sockets), so it's not a regression.

## Other observations (no action needed)

- **Caller-visible behavior change**: reading from a socket whose fd is already closed now raises `ValueError` up front instead of `ConnectionAbortedError` after a failed `SSL_read_ex`. Matches CPython's own behavior; SABnzbd treats both as reconnect triggers.
- Removing `dlclose` is correct — the `RTLD_NOLOAD` handle refcount pins a library `_ssl` holds anyway, and dropping it removes any risk of unloading code behind live function pointers. Resetting the `SSL_*` statics on partial failure fixes a real dangling-pointer case.
- `PyErr_Clear()` in the failure branch is right for the "linking is optional" design; previously a failed init could leave a pending exception to surface at a random later point.
- Nits: `SIZEOF_SOCKET_T` is defined but never used; `openssl_handle = NULL` in `cleanup` is a no-op on a local variable.
- CI: `3.15-dev` in the test matrix is right. Note the release job's cibuildwheel v4.1.0 won't produce cp315 wheels until 3.15 hits RC (or `CIBW_ENABLE: cpython-prerelease` is set) — fine for now, just something to remember at 3.15 release time.

## Verdict

Solid PR. The core — the versioned struct, the `PyWeakref_GetRef` migration, and the `openssl_init` cleanup — is correct and verified against CPython 3.10 through main. The inline comments below are small correctness/clarity fixes worth considering before merge.

src/unlocked_ssl.cc
```diff
@@ -164,7 +239,12 @@ static PyObject* unlocked_ssl_recv_into_impl(PySSLSocket *self, Py_ssize_t len,
     int retval;
     int sockstate;
     _PySSLError err;
-    PyObject *sock = GET_SOCKET(self);
+
+    PySocketSockObject *sock = NULL;
+    if (get_socket(self, &sock) < 0) {
+        PyErr_SetString(PyExc_ValueError, "Underlying socket connection gone");
+        return NULL;
+    }
```
Safihre, line 227: This `PyErr_SetString` clobbers whatever `get_socket()` produced: the real exception when `PyWeakref_GetRef` itself fails (contradicting the `// exception already set` comment inside `get_socket`), and the more precise `ValueError("Socket already closed")` for a closed fd.

Cleaner: make `get_socket()` responsible for setting the exception in all `-1` cases (set "Underlying socket connection gone" for the dead-weakref case there, like CPython's `get_socket` does) and have the caller just `return NULL`.

src/unlocked_ssl.cc
```diff
@@ -74,14 +99,60 @@ typedef enum {
...
+    PyObject *tmp = PyWeakref_GetObject(obj->Socket);
+
+    if (tmp == Py_None) {
+        return -1;
+    }
+
+    PySocketSockObject *sock = (PySocketSockObject *)tmp;
+
+    // Promote to an owned reference
+    Py_INCREF(sock);
```
Safihre: `PyWeakref_GetObject` returns NULL on error (non-weakref in `Socket`); then `tmp == Py_None` is false and `Py_INCREF(NULL)` crashes. Unlikely in practice since CPython always stores a weakref there, but the guard is one line: `if (!tmp || tmp == Py_None)`.

src/unlocked_ssl.cc
```diff
@@ -269,14 +344,24 @@ PyObject* unlocked_ssl_recv_into(PyObject* self, PyObject* args) {
         return NULL;
     }
 
+    if (!PyObject_TypeCheck(ssl_socket, SSLSocketType)) {
+        PyErr_Format(PyExc_TypeError, "argument 1 must be %s", SSLSocketType->tp_name);
+        return NULL;
+    }
+
     Py_ssl_socket = PyObject_GetAttrString(ssl_socket, "_sslobj");
```
Safihre, line 327: Pre-existing, but in the same hardening spirit as this PR: `_sslobj` is `None` on an `SSLSocket` that hasn't completed wrapping (or after `unwrap()`), and this cast eventually calls `SSL_read_ex` on garbage. An `if (Py_ssl_socket == Py_None)` → ValueError check would close the last obvious crash path here.

src/unlocked_ssl.cc
```diff
@@ -200,7 +272,9 @@ static PyObject* unlocked_ssl_recv_into_impl(PySSLSocket *self, Py_ssize_t len,
         } while (len > 0);
         err = _PySSL_errno(retval == 0, self->ssl, retval);
         Py_END_ALLOW_THREADS;
+#if PY_VERSION_HEX < SABCTOOLS_PY_HEX(3, 15)
         self->err = err;
+#endif
```
Safihre, line 257: Informational: on 3.15, `got_eof_error` is neither checked nor set here. CPython uses it so that, on OpenSSL 4, a read after an unexpected-EOF error still raises a clean `SSLEOFError` instead of a generic protocol error ([gh-148292](https://github.com/python/cpython/issues/148292)). Practical impact is nil since sabctools raises `ConnectionAbortedError` on all fatal errors and SABnzbd reconnects — but optionally set `self->got_eof_error = 1` on the fatal-error path for parity with the interpreter's state.

src/unlocked_ssl.cc
mnightingale: Removed, related to some future changes I may make

src/unlocked_ssl.cc
mnightingale: I want to keep the struct matching even though we don't use it.

Actually using it is an OpenSSL 4 "thing" but we aren't linking with it yet, so we will worry about it once it is an issue (fun)

mnightingale:
Alright I'm done now

Safihre:
Thanks 🎉

PR #190, opened 2026-07-09
Truncate rar3_s2k to max password length

mnightingale:


mnightingale:
fyi this is because I removed the password length restriction from SAB because rarfile 4.3 has it, but I didn't add it to our rar3_s2k implementation previously.

PR #191, opened 2026-07-10
rar3_s2k dont release gil

mnightingale:
There is no benefit to this, worse performance if anything.

### DIFF
--- a/src/utils.cc
+++ b/src/utils.cc
@@ -148,13 +148,11 @@ PyObject* rarfile_rar3_loop(PyObject *self, PyObject *args)
         unsigned long bufpos = nbytes & 63;
         nbytes += static_cast<unsigned long long>(seed_len);
         if (seed_len > 64) {
-            Py_BEGIN_ALLOW_THREADS
             Py_ssize_t dpos = 64 - static_cast<Py_ssize_t>(bufpos);
             while (dpos + 64 <= seed_len) {
                 rar3_corrupt_block(seed_ptr + dpos);
                 dpos += 64;
             }
-            Py_END_ALLOW_THREADS
         }
 
         const unsigned int x = base + j;


Safihre:
So why remove it?

mnightingale:
The loop runs 16384 times, it's releasing and reacquiring the GIL too much

> Remove it. There's no benefit here, and it's a small net cost.
> 
> `Py_BEGIN_ALLOW_THREADS` only helps when you release the GIL around a *substantial* chunk of pure-C work so other Python threads can run meanwhile. Neither condition that would make it worthwhile holds in this spot:
> 
> The wrapped work is tiny. `rar3_corrupt_block` is ~80 rounds of XOR/rotate over a 64-byte block, and the loop runs only `seed_len/64` times. For a typical RAR3 seed (password + 8-byte salt) `seed_len > 64` is usually false, so the block is often skipped entirely; when it does run it's a handful of microseconds.
> 
> You release and reacquire on every one of the 0x4000 (16384) outer iterations. Each release/reacquire saves and restores thread state and then contends for the GIL mutex on the way back in. Wrapping a few microseconds of work in that dance 16384× per call is pure overhead, and every reacquire is a point where you may have to wait for another thread.
> 
> There's no parallelism to gain anyway. The rest of the loop body — `PyObject_CallOneArg(update, ...)`, `digest()` — needs the GIL and dominates the runtime. Releasing it only for the microscopic corruption step means other threads get a vanishingly small window before this function grabs the GIL right back for the next `update`. The SHA1 calls serialize everything regardless.
> 
> So it's safe (the block touches no Python objects, which is why it doesn't crash) but not helpful. Drop the two macros and keep the plain loop:
> ```c
> if (seed_len > 64) {
>     Py_ssize_t dpos = 64 - static_cast<Py_ssize_t>(bufpos);
>     while (dpos + 64 <= seed_len) {
>         rar3_corrupt_block(seed_ptr + dpos);
>         dpos += 64;
>     }
> }
> ```
> The only structure where releasing the GIL would actually pay off is if you reimplemented the SHA1 `update`/`digest` in C too and wrapped the whole `0x4000` loop in a single `Py_BEGIN_ALLOW_THREADS`. As long as each iteration calls back into Python, keep the GIL held throughout.

I did try replacing Python calls with OpenSSL, maybe 10-20% faster but not worth more complicated build and requires maintenance for OpenSSL security updates so not worthwhile and naive SHA256 implementations are slower than just calling into Python which uses the OpenSSL optimised versions.

PR #199, opened 2026-07-31
Fix =yend CRC parsing, harden input validation and error handling

thezoggy:
- yenc.cc: search pcrc32/crc32 from start of =yend line; the old find(prefix, 5) skipped 5 chars after the prefix was already stripped, losing the CRC when it was the first field on the line
- yenc.cc: validate =ypart begin>=1 and end>=begin to prevent a negative part_begin and signed overflow in part_size
- yenc.cc: clamp expected buffer size before adding margin to avoid signed overflow on adversarial header values
- yenc.cc: align EncodingFormat values with sabctools.pyi (YENC=1, UU=2) so no enum member is falsy
- yenc.cc: propagate PyByteArray_Resize failure in Decoder_process; placement-new crc_expected; remove stray semicolon; fix stale comment
- yenc.h: drop unused <iostream> include, add <new>
- utils.cc: bytearray_malloc raises OverflowError/ValueError instead of passing unchecked sizes to PyByteArray_FromStringAndSize
- sparse.cc: check PyLong_AsLongLong error for the Windows handle and verify get_osfhandle_string creation
- sabctools.cc: check PyModule_Add* return values, use PyModule_AddObjectRef to avoid a leak on failure
- tests: regression test for =yend CRC field order; update bytearray_malloc expectations (ValueError/OverflowError)


mnightingale:
https://gist.github.com/mnightingale/cbd70624341d60fb2924c918f5e52dda should fix the build and a few warnings.

Not reviewed each change yet in detail, but generally it looks fine, a few annoying mistakes like EncodingFormat I intended to make start at 1 for the given reason. Some of the changes update to newer APIs were we had just not bothered with handling the extremely unlikely failure and bloat of code required but the newer APIs improve that.

PR #201, opened 2026-08-07
Vendor rapidyenc and build with scikit-build-core

mnightingale:
Replaces the hand-maintained `src/yencode/` and `src/crcutil-1.0/` with a pinned snapshot of [rapidyenc](https://github.com/animetosho/rapidyenc), built by its own upstream CMake.

Vendoring moves the expensive part upstream: rapidyenc owns the per-ISA flag matrix and its probes — which files get `-mavx2`, which get `-march=armv8-a+crc`, which probe must pass first — so new SIMD kernels arrive by re-vendoring rather than by hand.

It also unlocks par2cmdline-turbo, which we want to vendor next: the CMake side takes a project name and builds one nested project per call, so par2 should be another `add_vendored_project()` and a second script — not another change of this size. par2 benefits most, the single thread setup.py approach would take 20 minutes per test run and even more complicated flags.

- **`setup.py` → scikit-build-core + `CMakeLists.txt`**, because rapidyenc builds with CMake and driving it from a hand-rolled `build_ext` meant duplicating that flag matrix.
- **`tools/vendor_rapidyenc.py`** re-vendors from any tag/commit and pins the default, so a re-run reproduces the tree byte for byte; patches fail loudly rather than rotting. See `src/rapidyenc/VENDOR.md`.
- **Version is static in `pyproject.toml`** and compiled in from there, so metadata and `sabctools.version` cannot drift; publish refuses to release when the tag disagrees.
- **New `sabctools.crc_simd`** — the CRC32 kernel is chosen independently of the yEnc one and was invisible. A test fails if re-vendoring adds a kernel we do not name.
- **ccache on all platforms**, so builds and tests are faster: the vendored tree has no Python in it, so those objects are compiled once and reused for every interpreter in the matrix and between runs. Windows uses Ninja with MSVC to make that possible, as the Visual Studio generator ignores the compiler launcher.

---

I've tested the built wheels on macOS, Linux (x86 + ARM), and Windows.

We could add a `update_vendored.yml` to check for upstream changes and create a PR after calling `python tools/vendor_rapidyenc.py` I haven't for now but updates are not very frequent anyway.

The only fix here is possibly around which CRC method we use for a few calculations exposed to Python.


.github/actions/compiler-cache/action.yml
```diff
@@ -0,0 +1,157 @@
+name: Compiler cache
```
Safihre, line 1: What is the delay if we don't do all this caching and ninja stuff?

tests/test_kernels.py
```diff
@@ -0,0 +1,38 @@
...
+import re
+
+import pytest
+import sabctools
+
+ROOT = pathlib.Path(__file__).resolve().parent.parent
+VENDORED_HEADER = ROOT / "src" / "rapidyenc" / "rapidyenc.h"
+MODULE_SOURCE = ROOT / "src" / "sabctools.cc"
+
+ENCODE_DECODE_KERNELS = ("", "SSE2", "SSSE3", "AVX", "AVX2", "AVX512VL+VBMI2", "NEON", "RVV")
```
Safihre, line 12: Should we allow `""`? That would be very very strange if it cannot find anything. Same for below.

.github/actions/compiler-cache/action.yml
mnightingale: Admittedly the ccache + ninja benefit is almost nil with just rapidyenc, because it’s only ~30 source files.

**Total workflow duration**

| Run | Total duration | Sum of billable job-minutes |
|---|---|---|
| ccache + Ninja | 5m 0s | 29m 37s |
| No cache, no Ninja | 5m 9s | 27m 40s |

**Per-wheel build time**

| Platform | ccache + Ninja | No cache/Ninja | Delta |
|---|---|---|---|
| ubuntu-latest x86_64 | 3m 2s (182s) | 3m 44s (224s) | 42s faster (‑19%) |
| ubuntu-24.04-arm aarch64 | 2m 38s (158s) | 2m 53s (173s) | 15s faster (‑9%) |
| windows-11-arm | 1m 51s (111s) | 2m 4s (124s) | 13s faster (‑10%) |
| windows-latest | 1m 21s (81s) | 1m 28s (88s) | 7s faster (‑8%) |
| macos-latest | 51s | 51s | no difference |
| **Sum (wheels only)** | **9m 43s (583s)** | **11m 0s (660s)** | **77s faster (‑12%)** |

---

However, par2 is >130 source files so sees a much bigger difference

**Total workflow duration**

| Run | Total duration | Sum of billable job-minutes |
|---|---|---|
| No cache/Ninja | 10m 58s | 50m 50s |
| ccache + Ninja | 5m 59s | 36m 9s |

**Per-wheel build time**

| Platform | No cache/Ninja | ccache + Ninja | Delta |
|---|---|---|---|
| ubuntu-latest x86_64 | 9m 31s (571s) | 4m 31s (271s) | 300s faster (‑53%) |
| ubuntu-24.04-arm aarch64 | 5m 25s (325s) | 3m 49s (229s) | 96s faster (‑30%) |
| windows-latest | 2m 56s (176s) | 1m 40s (100s) | 76s faster (‑43%) |
| windows-11-arm | 3m 19s (199s) | 2m 43s (163s) | 36s faster (‑18%) |
| macos-latest | 1m 29s (89s) | 57s | 32s faster (‑36%) |
| **Sum (wheels only)** | **22m 40s (1360s)** | **13m 40s (820s)** | **540s faster (‑40%)** |

tests/test_kernels.py
mnightingale: `""` is a valid result, it's what rapidyenc could return as `RYKERN_GENERIC` when it couldn't find any compatible SIMD code to use, for example if it was ran on a RISC-V processor before the RVV implementation was added.

The first two tests in here are not particularly useful, but `test_every_vendored_kernel_is_named` is checking that after revendoring there is not a new `RYKERN_*` defined that we are not handling.

Safihre: We should at leas on Github CI not pass if it's empty.
Just as a sanity check.

